PR #109, opened 2015-01-13
add break 

redheli:
add break , so no need visit rest objects of this.visibleGeometry ,after find the target object


m-schuetz:
Thanks for the PR!

Could you please reassign the target from potree:master to potree:develop?

redheli:
@m-schuetz sure, I will create a new PR, close this one.

PR #110, opened 2015-01-16
add break

redheli:
add break , so no need visit rest objects of this.visibleGeometry ,after find the target object


m-schuetz:
Thanks!

PR #113, opened 2015-01-26
make examples/lion.html CPU friendly

luxigo:
- render scene when needed

Yeah i know it breaks the stats, but it saves electricity, batteries and knees

Keep it or trash it

BTW thanks for potree :-)


m-schuetz:
Interesting point. However, there are some issues with the current implementation. Only a part of the options invoke the necessary render/update calls.
For example:
- No new nodes are loaded unless you do some actions, like rotation or changing parameters
- Flight Navigation and Measurement tools should update as well

I think I'll generally leave CPU/battery concerns out for now because implementing something like that could cause a lot of headache. 

By the way, if you put anything online, let me know!

PR #115, opened 2015-02-02
Refs potree/PotreeConverter#57. POCLoader supports loading from subdirs.

chiccorusso:


m-schuetz:
see https://github.com/potree/PotreeConverter/issues/57

Thanks!

chiccorusso:
Wonderful! This implementation is really amazing. Big clouds appear in the blink of an eye!

m-schuetz:
You already tried this? That's crazy fast! 
I'd like to create a new master release by the end of the next week so if you find any problems, let me know.

chiccorusso:
Yes, I already tried it yesterday because I was waiting for it very greedily.
Just one thing: i see that the hrc files are a lot, maybe that will slow a bit down the streaming of the points, but that's a right price. I will send you some other screenshot and movies.

m-schuetz:
I guess a problem is the fixed depth of the hierarchy chunks. With a chunk depth of 5 and an octree depth of 10 or 11, a lot of hrc files are created at maximum octree depth where they aren't actually needed. Maybe hierarchy chunk depth should be set so that the last hrc files appear 4 or 5 levels before the leaf nodes.

PR #116, opened 2015-02-03
Fixes in the creation of the skybox.

chiccorusso:
The skybox does not work in WebGL contexts after the first one created. I often need to destroy and create new gl convas', and only the first one has the skybox.

This is beacause the current code modifies the value of THREE.ShaderLib["cube"], instead of creating a new copy each time.


potree:
Thanks for the PR!

PR #120, opened 2015-02-13
MeasuringTool considers element position

chiccorusso:
With this patch MeasuringTool picks the right points even in non full page canvas.


m-schuetz:
Thanks for the PR, works nicely!

I can't automatically merge this anymore so I've copied this change manually. 
If you do not commit potree.js and potree.min.js, chances are better that I can merge things automatically.
I'll also update the remaining tools.

m-schuetz:
I still have to do it for EarthControls

PR #122, opened 2015-02-15
Corrected typo in file_format doc

ghgr:
POSITION_CARTESIAN set to 32-bit signed integers, instead of 32-bit floats


m-schuetz:
thanks for the PR

m-schuetz:
I'll do some more updates to the documentation in the next week.

PR #132, opened 2015-03-15
Dispose invisible objects

michel-kraemer:
Hi Markus!

As discussed earlier I send you our proposal for memory management. The new function `disposeInvisible` disposes all objects that are not visible anymore. This is a very aggressive method as it does not care whether there is still enough free memory or not, but it makes sure memory does not overflow even for very large point clouds.

We tested the new function with a point cloud that is 14GB large and it works quite well. I hope you can merge the code or at least use it as a starting point to implement a more clever way to dispose objects. For example, you could collect objects to be disposed in a queue and only delete them after n frames or after m other nodes have been loaded or something like that.

Cheers,
Michel

### DIFF
--- a/build/js/potree.js
+++ b/build/js/potree.js
@@ -2467,6 +2467,46 @@ Potree.PointCloudOctree.prototype.update = function(camera, renderer){
 			this.updateVisibilityTexture();
 		}
 	}
+
+	this.disposeInvisible(this.children[0]);
+};
+
+Potree.PointCloudOctree.prototype.disposeInvisible = function(object) {
+	var stack = object.children.slice(0);
+	while(stack.length > 0){
+		var object = stack.shift();
+		if (object instanceof THREE.PointCloud && !object.visible) {
+			if (object.justReplaced) {
+				object.justReplaced = false;
+			} else {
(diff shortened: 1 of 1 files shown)


m-schuetz:
Hello Michel,

Thanks for the PR! It is a good starting point for me.
Rather than disposing all invisible nodes, I'd like to dispose least recently used nodes after a certain memory threshold has been reached. That way, nodes that might still be used often are kept in memory for future use.

michel-kraemer:
Yes, that would be great! Of course the threshold should be configurable. At least in our case we really need this aggressive type of garbage collection and we would set a really low threshold I guess.

Cheers,
Michel

m-schuetz:
I've implemented disposing least recently used nodes in the memory_management branch.
In the viewer, the load limit is set to two times the visible points limit: 
https://github.com/potree/potree/blob/develop/examples/viewer.html#L504

If you need immediate disposal of invisible nodes, you can set the pointLoadLimit to visible point limit.

```Potree.pointLoadLimit = pointCountTarget * 1 * 1000 * 1000;```

michel-kraemer:
Perfect. Thanks a lot!

PR #136, opened 2015-04-02
Fixes the pick in TransformationTool.

chiccorusso:


m-schuetz:
Already made it into the code a while ago. Thanks a lot!

PR #141, opened 2015-04-17
HTTP Basic Authentication

chiccorusso:
This patch allows to protect the cloud.js descriptor with HTTP Basic Authentication (only the cloud.js file, not point files or hrc).


m-schuetz:
Hello Federico,

Thanks for all the submissions and suggestions. I'll be back in a week and will work through them then!

adamsss92:
Hello, 
I'm not a javascript professionalist, so i've got a really silly question. How to apply this functionality in my .html? I would like to use "HTTP basic Auth", but i don't know how to handle this. I would like to sharemy potree point cloud with access for only specified user.

tiagodts:
> Hello,
> I'm not a javascript professionalist, so i've got a really silly question. How to apply this functionality in my .html? I would like to use "HTTP basic Auth", but i don't know how to handle this. I would like to sharemy potree point cloud with access for only specified user.

Did you figure this out?

PR #142, opened 2015-04-22
If CLIP_OUTSIDE, only visible pages are loaded.

chiccorusso:
With this patch, using CLIP_OUTSIDE, only the point files that intersect at least one of the active clip volumes are loaded.
In this way, the visible points are rendered in higher detail.

NOTICE: I am sorry I didn't revert the whitespace deletion in PointCloudMaterial.js.


m-schuetz:
I've now merged it into develop.
It ended up in Potree.js instead of PointCloudOctree.js since the visibility code has moved there.

Thanks for this nice PR, sorry I didn't get around to merging it sooner.

PR #184, opened 2015-11-09
Added support for presigned URLs

keattang:
- Added a function that joins URLs in a way that is aware of query strings.
- Used that function when requesting resources
- Tested and works locally and using a presigned cloudfront url


m-schuetz:
Hi, thanks for the PR!

Can you just give me an example of the problem this solves? 
Also, for merging please:
- Target potree:develop instread of master
- Don't use a formatter. Changes to formatting make it hard to grasp the actuall content of the commit.

keattang:
This solves the problem of having pointcloud data in a non public location, therefore requiring the URLs to the data to be presigned. For example, data in a secure S3 bucket being served via cloudfront

mv-:
We are experiencing this issue as well. We have point cloud data secured behind CloudFront and would like to access it via Potree. Is there a reason why support for this was never added to master? We can create a new PR for potree:develop if that's what it will take.

m-schuetz:
> Is there a reason why support for this was never added to master?  

* Target should be set to develop
* The diffs should only reflect the necessary changes, i.e. don't use a formatter or the whole file will show up in the diff which makes it difficult to find the actual changes.
* builds should not be part of a PR. I will do the builds myself, occasionally. 
* No merging conflicts would be helpful
* I'll need an example of why it doesn't work (what url is generated and why is it wrong) and how this is fixed (what does an URL with the change look like). I don't know anything about presigned URLs so a real world example to get an idea of it is necessary. 

Generally, the easier it is for me to analyze and validate the PR and its usefulness, the better, since handling PRs tends to be a lot of work.

manookie:
The most complete explanation of signed URL's can be found in Amazon's [documentation](http://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-creating-signed-url-canned-policy.html.).

I'll try and save you some reading and summarize. We store our point clouds in Amazon S3. When an S3 bucket is publicly accessible potree works as expected. An octreeDir set to `https://example.s3.amazonaws.com/dataset1` paired with the BinaryLoader accesses a node at `https://example.s3.amazonaws.com/dataset1/r.bin` without a problem.

Private S3 buckets require a little more work to gain access. Developers either need to use the AWS API to access resources directly or use a signed URL. Signed URL's generally work by including Expires, Signature, and Key-Pair-Id query string parameters when making a request. The value for Signature is a hashed, signed, base64-encoded version of a policy statement that’s created using a private key. The Signature in combination with the Key-Pair-Id, AKA public key, lets Amazon verify that a request for a resource came from an authorized/trusted source.

Due to the current implementation not being supportive or aware of query string parameters the following happens. 
`octreeDir:
https://example.s3.amazonaws.com/dataset1?Expires=13570344&Signature=nitfHrc39Dzz....&Key-Pair-Id=APKA9ONS7QCOWEXAMPLE`

`resulting url:
https://example.s3.amazonaws.com/dataset1?Expires=13570344&Signature=nitfHrc39Dzz....&Key-Pair-Id=APKA9ONS7QCOWEXAMPLE/r.bin`

The path to the resource is appended to the wrong part of the base URL. The desired behavior is that the code is a little smarter in how it generates its resource paths. Instead of blindly appending to the end of the base URL, check to see if there is a query string attached and insert the resource path at the correct position. 
`https://example.s3.amazonaws.com/dataset1/r.bin?Expires=13570344&Signature=nitfHrc39Dzz....&Key-Pair-Id=APKA9ONS7QCOWEXAMPLE`

We are currently working on a PR and once we've tested and validated that the changes work as expected it'll be out for review.

cherukumilli:
@martinheidegger @manookie 
We also need this functionality.
Can you please update us about the status of this PR?

keattang:
Hi all, sorry for the huge delay on this. Unfortunately I do not currently have the bandwidth to complete this PR. As such I am going to close it. 

The code changes that I made were very simple and could easily be replicated by those that need this. Please feel free to fork my branch and make the necessary changes to bring this PR up to scratch.

Apologies again for leaving this one hanging for so long.

manookie:
Check out my fork. I have to support an older version of the library in my
app but the changes are complete and working for 2.3 or 2.6. I can't
remember which one specifically. Unfortunately I just don't have the time
to get something ready for inclusion in the next release.

martinheidegger:
@cherukumilli all wir I do for this project I publish as soon as I have something. PRs can be labelled work-in-progress and I would have linked it immediately. Do you think it's acceptable to stress other contributors into investing their own spare time for your work?

@keattang even without a PR this issue seems valid. It should stay open until fixed, don't you think so?

@manookie this is the branch, right? https://github.com/manookie/potree/tree/support-for-presigned-urls-v1.3 might need a rebase ;)

manookie:
Yes sir, you've got the right branch.

cherukumilli:
@martinheidegger 
My apologies...the intent was not to stress other contributors.

keattang:
@martinheidegger happy to keep this open if you are.

jmprovencher:
@manookie Use signed cookies instead, this fixes your problem and does not require any change on Potree's side.

philip-firstorder:
I also need presigned s3 urls, any news here?

PR #191, opened 2015-12-09
Unify mouse controls between earthcontrols + orbit controls

chris-cooper:
This PR does the following...
* Unify mouse navigation between EarthControls and OrbitControls
* Flip the Y-axis on EarthControls rotation to match OrbitControls
* Allow ctrl-mouseleft fallback for rotation (e.g. mac, trackpad etc)
* Use mouse button names in code rather than numbers

Controls end up being...
* mouse left = move
* mouse middle OR ctrl + mouse left = rotate
* mouse wheel = zoom
* mouse right = zoom (orbitcontrols only)

### DIFF
--- a/src/EarthControls.js
+++ b/src/EarthControls.js
@@ -82,7 +82,7 @@ THREE.EarthControls = function ( camera, renderer, scene ) {
 			
 				var diff = mouseDelta.clone().multiplyScalar(delta);
 				diff.x *= 0.3;
-				diff.y *= 0.2;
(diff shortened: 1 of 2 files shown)


m-schuetz:
Hi, 

Thanks for the PR. Reversing the y-rotation direction for earth controls makes sense.
As for the remaining changes to behaviour, I'd rather leave them as they are.

When making a future PR, please set the target to the develop branch.

PR #194, opened 2015-12-16
flip y axis of orbit controls and make mouse buttons explicit in code

chris-cooper:
As per the comments on https://github.com/potree/potree/pull/191.  Three mouse constants are as per https://github.com/mrdoob/three.js/blob/317d6f66542ab63323c5dab6c7a4e9d08c07ca05/src/Three.js#L134


m-schuetz:
Thanks, looks good!

PR #196, opened 2015-12-17
Fix bug in pick function where other fields were not aligned

andyneff:
There appeared to be bug where instead of getting other fields for the correct
point, they were misaligned. By using hit.pIndex, I was able to get
the correct values in my tests

### DIFF
--- a/src/PointCloudOctree.js
+++ b/src/PointCloudOctree.js
@@ -887,11 +887,11 @@ Potree.PointCloudOctree.prototype.pick = function(renderer, camera, ray, params)
 				
 				}else{
 					if(values.itemSize === 1){
-						point[property] = values.array[i + j];
+						point[property] = values.array[hit.pIndex + j];
 					}else{
 						var value = [];
 						for(var j = 0; j < values.itemSize; j++){
-							value.push(values.array[i*values.itemSize + j]);
+							value.push(values.array[hit.pIndex*values.itemSize + j]);
 						}
 						point[property] = value;
 					}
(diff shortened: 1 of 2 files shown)


andyneff:
@m-schuetz Any thoughts/comments on this?

JWGmeligMeyling:
What do you exactly mean with fields? I've had some problems with the pick function giving off results (or at least, any tool using the pick function, such as the measurement tool or earth controls mouse intersection being off). I managed to fix these problems by setting the camera near value such as in the examples. (No idea why or how that works). Is this adressing the same issue or doing something different?

```js
			if(sg.radius > 50*1000){
				camera.near = 10;
			}else if(sg.radius > 10*1000){
				camera.near = 2;
			}else if(sg.radius > 1000){
				camera.near = 1;
			}else if(sg.radius > 100){
				camera.near = 0.5;
			}else{
				camera.near = 0.1;
			}
```

andyneff:
The pick function WAS picking the right point, that wasn't my issue.

I was trying to retrieve other information about the picked point (I guess they are called "materials" in the viewer: number of returns, normal values, etc... any of the additional fields put into the potree format from the original ply/las files) And what was happening was it randomly indexing into another point instead of the picked point. Since this was indexed differently than the location, it wasn't noticeable until I was accessing these other pieces of information and I knew the values were of a different point than I click (and was selected)

However, my PR is off of the master, and I am apparently supposed to be working off of develop, so I need to go retest all of this on develop when I find the time. (And I'll post my test ply file at that time to show you what I mean)

m-schuetz:
Thanks @andyneff, seems like I've fixed it in develop at some point, possibly after looking at this PR.
I'm going to close this here.

PR #204, opened 2016-01-18
Fixed 'PointCloudOctree.toTreeNode' function global variable issue

sanak:
After upgrading potree to latest "develop" branch, I encountered the following issue.
<pre>
Uncaught ReferenceError: pointcould is not defined                potree.js:5423
</pre>

https://github.com/potree/potree/blob/58ffe7fc6e83f8f8572c23f3a5a27a9afeb1481c/build/potree/potree.js#L5423

And "pointcloud" global variable seems to be used.
This fix will solve above issue.


m-schuetz:
Thank you for the fix!

Issue #205, opened 2016-01-21
MeasuringTool vertex can't be dragged in develop branch


sanak:
It seems not to work correctly after the following commit.
https://github.com/potree/potree/commit/08d7347d770c2861ccd6679ee26a049a83d42e7d

The following error messages were shown in Chrome Developer Tools.

<pre>
[GroupMarkerNotSet(crbug.com/242999)!:D03CD132E47F0000]GL ERROR :GL_INVALID_OPERATION : glDrawArrays: attempt to access out of range vertices in attribute 1
:
WebGL: too many errors, no more errors will be reported to the console for this context.
</pre>

m-schuetz:
Same as https://github.com/potree/potree/issues/202

Tried it with some of my point clouds but I can't reproduce the error. All of them work fine for me.
What parameters did you use for conversion and which version of PotreeConverter?

If you can upload an example or send me one (maybe downsampled with "-l 3") that would be helpful!

sanak:
Thanks for confirmation.
I also confirmed that the following environments were no problem (with examples/viewer.html).
- Windows 7 SP1 + Chrome
- Mac OSX (Yosemite) + Firefox

So, the cause seems to be Mac OSX Chrome specific WebGL issue.

davijo:
I get the same error on OSX Chrome (viewer+lion examples). Safari also stops working/picking after the first point picking though there is no informative messages related in the console.

I don't have the same issues (on OSX) with the latest dev branch on my local dev environment where Potree is used as a library.

sanak:
@davijo 
Thanks for information.
Are you using different version of three.js (r71) on your application ?
Because, I still encountered the problem with the latest "develop" branch.

martinheidegger:
This seems to be fixed. Closing the issue, please reopen if I am mistaken.

PR #206, opened 2016-01-22
Commented out classification shader material for #205

sanak:
I don't know well about WebGL shader,
but after commenting out the following part,
#205 issue didn't happen on Mac OSX + Chrome environment.

Please review this modification.
Thanks,


potree:
Hi @sanak,

I'm going to close this since things have changed a bit since then. Besides, this modification may be fine for specific mac environments but it removes a feature that works on other systems.

PR #225, opened 2016-04-19
Tools ui

monodo:
This PR:
- Make the tools accordion item open by default
- Make the profile window show up once the user inserts a marker
- Add support for named pointcloud. Warning: name must be defined in viewer.html
- Slightly cleans the code, but this is clearly work on progress

@m-schuetz please review


monodo:
Thanks!

PR #240, opened 2016-08-28
Propose alternative web server options for getting started

JamesMilnerUK:
Live server is a simple cross platform web-server that will automatically reload the page upon changes.


dkastl:
There are other options as well. Without installing anything (at least on Ubuntu) you can run a simple http server with Python:

```
python –m SimpleHTTPServer
```

PR #258, opened 2016-12-13
Add babel preprocessor to fix uglify errors

autra:
gulp-uglify do not support es6 and it makes `gulp build` fails on your develop branch. This commit adds babel preprocessing to fix this. 

Note that I also updated the build directory with the result of `gulp build` but it seems previous commits did not do so. I can remove this part of the commit if you like


autra:
pinging @m-schuetz :-)

m-schuetz:
Hi,

If I understand this correctly, babel takes ES6y code and transform it to something pre-ES6, which is then minified?
I'd rather not have to do that and find an ES6 compatible minifier, instead. Do you, by chance, know of one?

m-schuetz:
Okay, it would seem like there isn't.

> Note that I also updated the build directory with the result of gulp build but it seems previous commits did not do so. I can remove this part of the commit if you like

That would be great! I do builds every now and then to keep the diffs a bit cleaner.

autra:
> If I understand this correctly, babel takes ES6y code and transform it to something pre-ES6, which is then minified?

Yes it does. Which allow a better cross-browser experience too (but by using three.js, potree won't work in IE<11 anyway, so maybe we don't care).

So to summarize, I see 3 possible solutions:
- use babel, but only with [es2015-block-scoping](https://babeljs.io/docs/plugins/transform-es2015-block-scoping/) plugin (will change const and let to var I guess) to minimize the impact.
- use another bundler, like webpack or browserify *if* they use their own minifier (which I'm not even sure of) and not uglifyjs2. But that would be a big change in your build system. 
- Or simply drop uglification until uglifyfs2 (and gulp-uglify) supports what you need. Anyone who wants network perf in production would activate gzip on their server, making minifying less important I guess. But still, `gulp build` with minifying disabled yields a 1.52MB potree.js in your develop branch... 

So yeah, no ideal solution :-/ Please let me know which solution you prefer and I'll amend this.

m-schuetz:
Thanks for the infos. 

> But still, gulp build with minifying disabled yields a 1.52MB potree.js in your develop branch...

It seems like majority of the size (~1mb) stems from a base64 encoded WebWorker for greyhound which won't get smaller through minfication. I guess the best size improvement would be to load this code on demand, rather than shipping it by default.
I'll give the minifying part another thought when this is done.

m-schuetz:
laslaz, greyhound and binary decorder workers are now loaded on demand. 
potree.js is now 480kb instead of ~1.5mb.

see: https://github.com/potree/potree/commit/8b5ffb77884ae1759dbc581fe877d36906bed030

I think I'll leave it at that for now; 480kb sounds reasonable. 
Thanks for bringing this to my attention.

autra:
Thanks @m-schuetz !

PR #283, opened 2017-02-10
Simplify scale/offset usage for Greyhound usage.

connormanning:
This does some simplification to the Greyhound-loader's scale/offset usage.  Also, in the sample viewer for Greyhound, resources are updated to current names, and resources can be overridden with query parameters.

### DIFF
--- a/examples/entwine_greyhound.html
+++ b/examples/entwine_greyhound.html
@@ -14,9 +14,9 @@
   </head>
 
   <body>
-  
+
 	<script src="../libs/jquery/jquery-3.1.1.js"></script>
-	
+
 	<!--<script src="../libs/other/webgl-debug.js"></script>-->
 	<script src="../libs/perfect-scrollbar/js/perfect-scrollbar.jquery.js"></script>
 	<script src="../libs/jquery-ui/jquery-ui.min.js"></script>
@@ -28,9 +28,9 @@
 	<script src="../libs/proj4/proj4.js"></script>
 	<script src="../libs/openlayers3/ol.js"></script>
     <script src="../libs/i18next/i18next.js"></script>
-	
+
 	<script src="../build/potree/potree.js"></script>
-	
+
 	<!-- INCLUDE ADDITIONAL DEPENDENCIES HERE -->
 	<!-- INCLUDE SETTINGS HERE -->
 
@@ -54,11 +54,11 @@
     </div>
 
 	<script>
-	
+
 		window.viewer = new Potree.Viewer(document.getElementById("potree_render_area"));
-		
+
 		viewer.setEDLEnabled(true);
-		viewer.setPointSize(1);
+		viewer.setPointSize(0.3);
 		viewer.setMaterial("RGB");
 		viewer.setFOV(60);
 		viewer.setPointSizing("Adaptive");
@@ -67,25 +67,50 @@
 		viewer.setIntensityRange(0, 300);
 		viewer.setWeightClassification(1);
 		viewer.loadSettingsFromURL();
-		
+
 		viewer.setDescription('Streaming point clouds from an <a href="https://entwine.io/" target="_blank">Entwine</a> backend.');
-		
+
 		viewer.loadGUI(() => {
 			viewer.setLanguage('en');
 			$("#menu_appearance").next().show();
 			//viewer.toggleSidebar();
 		});
-		
-		{ 
-		
-			//var server = "greyhound://data.greyhound.io/resource/";
-			//var resource = "autzen-h";
-			
-			//var server = "greyhound://data.greyhound.io/resource/";
-			//var resource = "nyc-h";
-			
-			var server = "greyhound://data.greyhound.io/resource/";
-			var resource = "redrock-h";
+
+        var getQueryParam = function(name) {
+            name = name.replace(/[\[\]]/g, "\\$&");
+            var regex = new RegExp("[?&]" + name + "(=([^&#]*)|&|#|$)"),
+                results = regex.exec(window.location.href);
+            if (!results || !results[2]) return null;
+            return decodeURIComponent(results[2].replace(/\+/g, " "));
+        }
+
+		{
+			var server = "greyhound://cache.greyhound.io/resource/";
+
+            /* Some current public resources:
+                autzen
+                half-dome
+                iowa
+                iowa-bridge
+                iowa-city
+                lake-isabella
+                lone-star
+                nyc
+                nz
+                red-rocks
+                sncf
+                space-shuttle
+                st-helens
+                vanuatu-village
+            */
+            var resource = "red-rocks";
+
+            if (getQueryParam("server")) {
+                server = getQueryParam("server");
+            }
+            if (getQueryParam("resource")) {
+                resource = getQueryParam("resource");
+            }
 
 			Potree.loadPointCloud(server + resource + "/", "autzen", (e) => {
 				viewer.scene.addPointCloud(e.pointcloud);
@@ -93,9 +118,9 @@
 				viewer.scene.camera.near = 10;
 			});
 		}
-		
+
 	</script>
-	
-	
+
+
   </body>
 </html>
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -157,10 +157,10 @@ gulp.task("scripts", ['workers','shaders'], function(){
 
 	gulp.src(paths.html)
 		.pipe(gulp.dest('build/potree'));
-		
+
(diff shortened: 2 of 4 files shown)


m-schuetz:
Hi Connor,

I'm getting stutter issues with this commit. 
Coordinate values are no longer translated to the origin which results in huge coordinate values in a Float32Buffer which has rather limited precision. Additionally, the camera position is now twice of what it was before.

before:

    viewer.scene.pointclouds[0].root.geometryNode.geometry.attributes.position.array 
    //234.6300048828125, 62.5, 476.739990234375
   // viewer.scene.pointclouds[0].position will be added to these coordinates to retrieve the original coordinate values 

after:

    viewer.scene.pointclouds[0].root.geometryNode.geometry.attributes.position.array 
    // 482184.625, 4390222.5, 1946.739990234375
    // viewer.scene.pointclouds[0].position will be added and now we get about 2x the original value

I'd like to make 1.5RC public later today or tomorrow, would be nice if we can fix this now! 
I guess I can take the entwine example html without issues. What are the changes in the GreyhoundLoader and geometry for?

connormanning:
The original issue I set out to solve was that the `nyc` dataset was not loading to its full depth.  I saw some `hierarchy` requests that I believe had an incorrect local scale/offset, since they were all returning `null` when there should have been points at that depth.  Trying to debug, I had trouble with the differing scale/offset values for each node request.

Since the entire dataset fits within an `int32` when centered around the origin and scaled at .01, the entire dataset should be queryable with no precision loss with a single transformed coordinate system.  So you should be able to use a scale of `.01`, and an offset of `[-8235590, 4969360, 340]` and perform the tree traversal based on a transformed bounds of `[-3152000,-3152000,-3152000,3152000,3152000,3152000]` without worrying about changing your reference frame for each node, just pass the `scale` and `offset` values with each Greyhound call.

For this reason the complexity of the per-node local scale/offset seemed unnecessary to me, so I tried to use only a single one.  However I didn't realize I had translated the coordinate space of the result going down to the renderer, my intention was to keep the end result the same - centered around the origin.

I'm looking at the code now trying to resolve it - does what I'm saying make some sense or am I making any bad assumptions?

m-schuetz:
I see, had the same problem figuring out how to handle greyhound coordinate scaling/transforming and ended up producing some confusing stuff.

Anyway, I think this here is the issue:

image


The offset will end up being the pointcloud.position and it is the value that is subtracted from all coordiantes before they're stored in the Float32Array, as far as I remember.

connormanning:
Interesting.  Ok still looking - however my interest was in these lines though: https://github.com/connormanning/potree/blob/develop/src/workers/GreyhoundBinaryDecoderWorker.js#L117-L119

Since, after my PR, the points are all coming in pre-scaled/pre-offset to the reference frame of `[-3152000,-3152000,-3152000,3152000,3152000,3152000]`, that those lines are moving the points back into their global coordinate space rather than centered around the origin.

m-schuetz:
Could you also revert and not commit the build folde? Makes looking at diffs and merging easier.

connormanning:
Done ^

connormanning:
I'm not quite certain how the scale/offset values are used internally.  Currently with this PR, the binary point data is coming back with the scale/offset values *already* applied.  I know that somewhere in the library this needs to be accounted for, since that scale/offset will be needed by the map and the tools so that their values are in native coordinate space.  However since the binary data from Greyhound is already scaled/offset, there should be no computation needed there - so XYZ should be left alone from the binary result.

So should I set `pgg.scale = 1` and `pgg.offset = [0,0,0]` so that no additional computation is performed on the points?  If so, how do I notify Potree of the scale/offset so that the web tools can use the correct coordinate space?

m-schuetz:
> I'm not quite certain how the scale/offset values are used internally 

I wish I could claim otherwise. I'm going to debug a bit but I've got to go soon so maybe tomorrow if I can't make it now,

m-schuetz:
If the coordinates from greyhound are already translated to the origin, then the + nodeOffset[0] shouldn't be done.
https://github.com/connormanning/potree/blob/develop/src/workers/GreyhoundBinaryDecoderWorker.js#L117

The bounding box of the node should have a minimum of (0, 0, 0) and the maximum is equal to the size of the node. The offset of the node is the value that is going to be added to it's location during rendering. 
So I believe that this part:
https://github.com/connormanning/potree/blob/develop/src/loader/GreyhoundLoader.js#L164
should be reverted to what it was before:

    var offset = boundingBox.min.clone();
    boundingBox.max.sub(boundingBox.min);
    boundingBox.min.set(0, 0, 0);

The getUrl function (https://github.com/connormanning/potree/blob/develop/src/PointCloudGreyhoundGeometryNode.js#L67 ) later uses the boundingBox and the offset to generate the query. This has to be adapted under the assumption that offset is the minimum of the world space bounding box, boundingBox.min on the other hand is (0, 0, 0) and boundingBox.max is the size of the node.

I'll have to go now so I can't test/verify this right now but I can check tomorrow. I'll defenitely need your help in generating the url, though. What should the url look like if

offset: 481950, 4390160, 1470
boundingBox.max: 940, 940, 940 
scale: 0.01 ? 

Scale is the coordinate precision so 0.01 = cm precision.

m-schuetz:
I'll need to refactor that whole thing at some point...

connormanning:
Hmm trying to understand the flow... I didn't realize that the bounding box of the node should always have a min of `[0,0,0]`.  So I was misinterpreting the offset of the node.  I was considering the node's offset to be the offset necessary to transform into global space.  But really it is an offset to transform from `[0,0,0]` into the local coordinate space, which is offset from the global coordinate space.  If I understand correctly.

So for what I did, the node's scale/offset should have been `1` and `0` since they were already in transformed coordinate space, I believe.

For your URL, I think you'd just have to add your local offset to the global offset from the minimum of `greyhoundInfo.bounds`.  I've been messing around with this now, so hopefully I'll have a fixed version at some point.

m-schuetz:
It's grown like that at some point. During loading, point cloud coordinates are translated to the origin by the min of the bounding box. That's why the bounding box min ends up being 0. Point coordinatez and bounding box are then in the range 0 to size. The offset is used the move everything back where it belongs during rendering and pointcloud.position takes on the value of offset. Scale is the value used to scale coordinates from fixed precision integers to floating point coordinates. Hope this helps. 

Some things in there don't make much sense because the handling of large coordinates has changed a few times. 
-- 
Sent from my Android phone with GMX Mail. Please excuse my brevity.On 10/02/2017, 18:45 Connor Manning <notifications@github.com> wrote:

  
    
      Hmm trying to understand the flow... I didn't realize that the bounding box of the node should always have a min of [0,0,0]. So I was misinterpreting the offset of the node. I was considering the node's offset to be the offset necessary to transform into global space. But really it is an offset to transform from [0,0,0] into the local coordinate space, which is offset from the global coordinate space. If I understand correctly.
    
    
      So for what I did, the node's scale/offset should have been 1 and 0 since they were already in transformed coordinate space, I believe.
    
    
      For your URL, I think you'd just have to add your local offset to the global offset from the minimum of greyhoundInfo.bounds. I've been messing around with this now, so hopefully I'll have a fixed version at some point.

connormanning:
Seems reasonable enough, but some of my assumptions were wrong. I'll try to fix my changes today.

connormanning:
I've force-pushed over the original PR with a new one.  It's a much smaller change now that I understand what's happening a little better.  The only real change I've made is that the scale/offset tranformations sent to Greyhound were calculated differently for `hierarchy` requests than they were for `read` requests.  I've consolidated those transformations, using the same transformation for `hierarchy` as `read`, and it seems to be loading the `nyc` data to full resolution.

There seems to be a bit of a twinkling effect if you zoom in all the way and then rotate the point cloud, maybe some form of Z-fighting or something related to the occlusion determination within the renderer?  However I don't think that's necessarily relevant to this PR so I may look into that separately.

m-schuetz:
Looking good! Thanks for this PR and the entwine support in general!


> There seems to be a bit of a twinkling effect if you zoom in all the way and then rotate the point cloud, maybe some form of Z-fighting or something related to the occlusion determination within the renderer? 

I'm not sure what you're referering to, maybe our GPUs produce different results here. z-fighting could be possible, since I'm now setting the near clip plane depending on how far you're zoomed in. It's to avoid z-fighting whole zoomed out and clipping too many points while zoomed in. So if you're zoomed in, the near clip plane is small which could lead to z-fighting at distant regions. 
Not experiencing it myself, though.

PR #284, opened 2017-02-11
Normalize color/intensity if needed for Greyhound resources

connormanning:
This builds upon #283 for the Greyhound loader.  I've submitted it separately for tracking purposes, so if #283 gets nixed I will force-push over this one with these changes isolated.  The changes here are quite small but looks a bit large since I pulled the HTTP fetches into functions toward the top.

The basic idea is to check a few thousand points at the base depth, and if any of them have color or intensity values > 255, then those values will be truncated later within the decoder worker.

Some sample resources where this is necessary are `sncf` and `dk`.  For example you could use [this link locally](http://localhost:8080/examples/entwine_greyhound.html?resource=sncf), which uses 16-bit color and is all white without this PR.

### DIFF
--- a/examples/entwine_greyhound.html
+++ b/examples/entwine_greyhound.html
@@ -14,9 +14,9 @@
   </head>
 
   <body>
-  
+
 	<script src="../libs/jquery/jquery-3.1.1.js"></script>
-	
+
 	<!--<script src="../libs/other/webgl-debug.js"></script>-->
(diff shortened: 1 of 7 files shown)


m-schuetz:
Thanks for the PR! 
I'm going to remove the intensity normalization since that's whats viewer.setIntensityRange(min, max) is intended for, e.g. viewer.setIntensityRange(0, 255)  or viewer.setIntensityRange(0, 65535).

Additionally, you can also adjust the intensity range in the UI

image

PR #309, opened 2017-03-20
Add let keyword before initSettings function declaration

pierotofy:
As the title says. I could not integrate potree in a project without adding it.


m-schuetz:
Makes sense, thanks for the PR.

PR #328, opened 2017-05-03
Fix Greyhound intensity normalization detection

elemoine:
This fixes the Greyhound code that detects when the intensity should be normalized. This is following up on https://github.com/potree/potree/commit/3fb45b6b26d4f3fb4b49d513fdd35757a2b08a80#commitcomment-21990352.


potree:
Does this look fine for you, @connormanning ?

connormanning:
👍

PR #376, opened 2017-08-14
Documentation for the file format.

martinheidegger:
This is supposed to be the first stab at a documentation for the potree file format. I added it to this repository because I saw it the fittest but I am still torn about it. It basically describes the things I could gather from reading the source code. I have skipped quite a cases that are still supported from the old potree file formats.


potree:
I fixed some parts of the descriptions. Looking good so far, except the mask example which I don't get:


image

martinheidegger:
I was brooding about the mask example quite a bit when I initially wrote it, simply because I am not sure how to document it best. I just commited a few formatting improvements, maybe its better readable now?

m-schuetz:
Thanks!

PR #379, opened 2017-08-14
Updated eslint and fixed linting

martinheidegger:
Eslint has had a version update that fixes a bug which was visible in the repo through a linting error. Since NPM on travis uses a version of NPM that doesn't support `package-lock.json` it might have fallen through the gaps during development.  This PR updates eslint and fixes the linting error.


potree:
Cool, thanks!

PR #385, opened 2017-08-21
Create issue template

martinheidegger:
Added template questions that hopefully results in nicer issues.


dkastl:
@martinheidegger would it make sense to ask for the version of Potree, that causes the issue?

martinheidegger:
@dkastl Added a mention of the version.

Issue #211, opened 2016-02-26
potreeViewer on webpack/nodejs


qdnguyen:
Hi, 

Do you think a potreeViewer version on nodejs with webpack is necessary? 

In Paris CodeSprint 2016, i tried to put potreeViewer code into AMD context (it also work with latest version of threejs), then we can create a webpack project to run it on nodejs. 

If you think it's necessary to create a branch for it. Please let me know.

mbredif:
That would be great and increase the user base of potree as potree could then be made available on npm.js, just as three.js already is.
The use case of @qdnguyen  is integrating potree inside another three/webgl viewer that we are developing (https://github.com/iTowns/itowns2), in order to delegate the point cloud visualisation to a potree layer. There, providing potree as a modular npm package would help a lot. 
@qdnguyen, could you prepare a PR with what you have coded so far to make the proposal more concrete?

dkastl:
To make Potree a modular library is also in my interest. I think every project comes to the point, where it makes sense to think of a new major release, that allows to break compatibility with previous releases.

I'm interested in opinions or proposals of best practices how to create a modern and modular Javascript library. Sharing the work among more people would be also great.

monodo:
As the [viewer](https://github.com/potree/potree/tree/develop/src/viewer) in Potree now integrates most functionalities that where implemented in [PotreeViewer](https://github.com/PotreeViewer/PotreeViewer) I'dont think will continue working on PotreeViewer. 

:+1: for the creation of a modern and modular Javascript library that would allow, for instance, to create a specific build containg only the parts required for the profile tool integration in webmapping applications

martinheidegger:
As far as I can tell, there is no "one-size-fits-all" approach to frontend-code on NPM.   Let me list the issues I can see before it would be reasonable to publish `potree` on NPM:

1. All the dependencies (currently stored in the [`libs`](https://github.com/potree/potree/tree/develop/libs)) folder need to checked if they are hosted on npm already _(in the correct version)_ or not.
[...]

PR #391, opened 2017-08-30
Move codebase to commonjs modules.

### LINKED ISSUE #369: Use Browserify to package deployments
Currently the build process for creating the `potree.js` simply combines a list of files: ([gulpfile#L147](https://github.com/potree/potree/blob/5e67da40419aa9b6eb3dad4a640ef4d613d773fc/gulpfile.js#L147)). I think it would be a good idea to move to  [`browserify`](https://www.npmjs.com/package/browserify). It would be beneficial because we wouldn't need to maintain the list of files that are in the `potree.js`; The order in which the files are combined would always be correct; and it would act as a good preparation to deploy `potree.js` on npm.

martinheidegger:
The current code builds the final `potree.js` by [concatenating files](https://github.com/potree/potree/blob/c057600e3acdbb5e0a36e3fb6c3e971b6636dfcc/gulpfile.js#L146-L149). This process comes with several problems:

1. It is hard to figure out what code depends on each other, which is difficult when trying to do any refactoring like necessary for #388 . It also partly solves the problem that some classes/methods are not easy to find because the name <-> file relationship is not given. _(Sidenote: It also reduces some file lines significantly)_
2. By not having internal dependencies, external dependencies are also not clearly defined. Making it hard to publish it to npm #211 
3. The current code concatenation does not support minifying or [sourcemaps](https://www.html5rocks.com/en/tutorials/developertools/sourcemaps/) which makes the code big and hard to read as well.
4. By using the global dependencies such as `Three.js` or `jQuery` it is not possible to have a different version of Three.js running on the same page.
5. By having clear dependencies everywhere its a lot easier to write tests. Dependencies make it easy to [mock](https://stackoverflow.com/questions/6679914/what-javascript-mocking-frameworks-are-people-using).

In this PR I decided to move to [Common.JS](http://www.commonjs.org/specs/modules/1.0/) modules, which are the base of Node.js and have a lot of build tools like [browserify](https://browserify.org) and because they are the core system for distribution in NPM. I decided against [require.js](https://requirejs.org) or [amd](https://github.com/amdjs/amdjs-api/wiki/AMD) modules because they seem to be barely used by the community and because I couldn't find anything for workers. I decided against [es6 modules](http://2ality.com/2014/09/es6-modules-final.html) simply because the syntax overhead is higher in es6 modules which increases the risks for merge errors towards other code also written now, and also because I am not sure if I can make it compatible with the current code base _(also the worker situation is a bit dire)_. I decided against immediately moving everything to TypeScript modules because, again, the merges would become even more difficult.

Tasks:

- [x] Moving shaders to browserify system.
     _Shaders are currently done using [`browserify-shader`](https://www.npmjs.com/package/browserify-shader) which will should allow follow-up PR's to simplify the shader configuration_
- [x] Moved globals into a placeholder module `./src/context.js`.
     _Globals such as `Potree.version` or `Potree.scriptPath` are dependencies to configuration variables. Any reference to such a global variable would require me to require `Potree` itself (circular ...). Having a separate context file made my work easier and should help in future improvements._
- [x] Move extensions to utils
      _Extensions modify loaded modules (namely three.js) which stands in general violation of the commonjs/es-modules contract._
- [x] Make sure that in each file there is only one module.
      _Multiple classes/functions per file make it hard to understand the relationship between code (particularly `utils`)_
- [x] Add browserify to build system.
- [x] Make sure all dependencies are fed through commonjs modules.
    - [x] Update `d3` from version 3 to 4 because 3 is not commonjs compatible
    - [x] `jquery` and `i18n` are fed through placeholder modules.
          _(integration through commonjs was painful, but without placeholders its impossible to track what files use jquery/i18n)_
- [x] Rebase against #402 _(#404 respectively)_
- [x] Think about how to deal with workers.
~~[ ] Give placeholders for shaders.~~
- [x] Use the css files from the `node_modules` dependencies.
    _In order for the css files to have the same version as the javascript dependencies, the examples need to change their point of reference_.
- [x] Add the package.json dependencies to the build.
- [x] Move other gulp scripts to use the browserify modules.
- [x] Test if everything still works.
- [x] Remove obsolete files/libraries.
- [x] Remove the old build system.
- [ ] Pull the files that back together in a structure resembling what was there before.

Closes #369.
In preparation for: #211

_Note: This PR also removes the last references to a global `viewer` object and any bug that may be caused because of that._
_Note: This PR fixes `viewer.setClipMode`._


m-schuetz:
File structure should remain mostly the same; methods should not be moved to seperate files, each.
Only exception I can think of are larger classes, but not small stuff like simple enumerations.

martinheidegger:
Dependencies in _(any)_ module systems need have declarations in order to find out which belongs where. By pulling the utils apart in files I got a better, clearer picture of what dependency is needed where because I needed to look over less code. Since smaller files will in future make debugging, merging and refactoring easier I thought it would be good to just keep them that way. I am sorry that you dislike the way I prepared this. Your request to merge the files together - in the same structure as before - this will probably take me a few days more.

m-schuetz:
Just wondering, what exactly makes it an issue? Circular dependencies?

martinheidegger:
There is no "issue". Just a hasty decision on my side. I realized while adding the require statements that I had to check every line of code to make sure that the dependencies are written in the require statements. Since many statements had just `Potree.<something>` or `Potree.util.<something>` mentioned but the actual references could have been in any file it was hard for me to tell what references would actually be required. After pulling the structure apart it was a lot easier to know which dependency was actually necessary.

martinheidegger:
I just updated my previous comment as I realized that it was near unreadable. If I may ask the question: I only see advantages in having the files apart as they are now, why would you want them back together? _(files are cheap, debugging is hard)_

dkastl:
I also think, that having more but shorter files is an advantage and makes it a lot easier to work with the code. In the past I have written code with thousands of lines in one file, and I would never do this again ... more than enough lessons learned ;-)  

The only real disadvantage of splitting code into smaller files is, that you need to care of dependencies, which is hard without the right tools. But that's actually what this PR aims for: to use the commonjs and other state-of-the-art tools, so this shouldn't cause any extra work.

There is probably a little time necessary to get familiar with this, but I'm pretty sure it will pay off very soon.

adrelino:
The work @martinheidegger has done lately about splitting up the code into small files allowed me to easily extract only the required components of Potree into the core branch in my fork: https://github.com/adrelino/potree/tree/core

In particular, I removed the GUI from the Viewer because l will later develop a custom GUI using Angular 4 to interact with Potree, which will have only a very limited subset of the current GUI's capabilities and a different styling.

This allowed to reduce the dependencies in [package.json](https://github.com/adrelino/potree/blob/core/package.json) to a minimum, which are all available via npm:
```javascript
 "dependencies": {
    "@tweenjs/tween.js": "^17.1.0",
    "bluebird": "^3.5.0",
    "jquery": "^3.2.1",
    "three": "^0.87.1"
  },
```

Running
`npm run browserify`
then creates a much smaller potree.js file including only the absolutely needed dependencies.

I adapted [examples/viewer.html](https://github.com/adrelino/potree/blob/core/examples/viewer.html), now one only has to include one css and one js file:
```html
<link rel="stylesheet" type="text/css" href="../build/potree/potree.css">
<script src="../build/potree/potree.js"></script>
```

The live demo of examples/viewer.html can be seen here:
https://rawgit.com/adrelino/potree/core/examples/viewer.html

So in summary, I see a great benefit in his work and hope that this PR, albeit its huge changes, will get merged and be used as basis for all further development.

m-schuetz:
It's up for debate how certain utility function could be restructured but moving very small function into seperate files is a way too fine grained seperation, e.g. files like:

- https://github.com/martinheidegger/potree/blob/browserify/src/utils/normalizeUrl.js (5 lines)
- https://github.com/martinheidegger/potree/blob/browserify/src/utils/pathExists.js (9 lines)
- https://github.com/martinheidegger/potree/blob/browserify/src/utils/projectedRadius.js (6 lines)
- https://github.com/martinheidegger/potree/blob/browserify/src/utils/pixelsArrayToImage.js (23 lines)
- https://github.com/martinheidegger/potree/blob/browserify/src/utils/rightView.js (7 lines)
- https://github.com/martinheidegger/potree/blob/browserify/src/utils/topView.js (7 lines)

and many many more files like this. These functions do not warrant their own files. They should be grouped into certain helper classes/namespaces/whatever. Keeping the previous structure might be the easiest thing to do but if you have good suggestions about relocating some functions into other helper classes that's fine as well, as long as we don't end up with a one file per method solution.

m-schuetz:
What is the rationale for these files?

- https://github.com/martinheidegger/potree/blob/browserify/src/jquery.js
- https://github.com/martinheidegger/potree/blob/browserify/src/context.js
- https://github.com/martinheidegger/potree/blob/browserify/src/i18n.js
- https://github.com/martinheidegger/potree/blob/browserify/src/index.js

martinheidegger:
It is way more work for me trying to renegotiate the place for every method than trying to put it in the old structure. Either I leave it as is now and restructurings come in future PRs or I put it in the old place.

I updated the description of this PR to contain the information about files such as "context" and i18n. The index file contains the reasoning in the header. It seems like you didn't see that. Can you be more specific: what needs clarification?

DanielSmon:
Hi guys, and thanks @martinheidegger for your efforts. I'm excited to see this work as we were looking at doing something similar to make it easier to integrate Potree with our own project. One point of difference is that we're using webpack. We are also using TypeScript, so a big +1 from us to move in that direction in the future, and we're keen to contribute.

I have yet to read through this thread and the associated commits in detail, but in short, I'm in favour of separating out utility functions in to small files. I agree that it makes it simpler to get to know the codebase coming from the perspective of someone that is new to it. Tooling (such as VSCode) makes it easy to work with separate files as well.

adrelino:
@martinheidegger what is the current status? I checked your latest commits (after the rebase) but the library is somehow broken now. Do you need help e.g. with workers?

martinheidegger:
@adrelino I am working on the workers _(got disrupted by other tasks, on it again!)_

martinheidegger:
I figured out most of the details regarding workers and external dependencies. I also tested the examples in the current state. The build system still needs to be put back in place but I think the heavy stuff is out of the way.

martinheidegger:
I progressed quite a bit in the gulp scripts, though I found some errors at the last check. Will need to look over things once more. In any case the current structure is how I plan to get a review for the PR (once I fixed the bugs). @m-schuetz Because of some discussion going on in the comments: Is it okay if the PR contains lots of small files or do you need the back at the old place (just to make sure).

m-schuetz:
I'd still like to avoid 5 liners but I could probably reassemble them myself afterwards. Or do you see any issues arising if I just merge some files together?

m-schuetz:
Let me know when it's ready to test and how to do it. (build, watch?, etc)

martinheidegger:
I still have one issue to clean up which is that the laslaz and dem worker isn't built properly. In will have to look at it again with a clear head.
But you could check out the other parts, basically I am testing and cleaning at this point. The scripts are available now through gulp as well as through npm (i.e.: there is `npm run build` and `gulp build`) the reason being that there are people who don't have _(and want)_ gulp installed globally and this way it works as well.

```text
npm run build ... builds all there is to build (copy, scripts, etc.)
npm run watch:* ... watches all watchable tasks
npm run script:* ... builds all scripts
npm run min:script:* ... build all scripts and uglifies them
npm run copy:* ... copies everything to copy
npm run script:<--- enter script id here ---> ... builds one particular script (main, bin, dem, gre, las laz are available) i.e. "npm run script:main"
npm run watch:<--- enter build task here ---> ... watches the changes of the task section. i.e. "npm run watch:min:script:*" or "npm run watch:copy:license"
npm run clean ... removes the build folder
npm run release ... clean build basically
npm run webserver ... runs a webserver in the current directory
npm run liveserver ... runs a liveserver in the current directory that automatically refreshes when a file changes
npm run watch+liveserver ... runs the liveserver while also building when any file changes
npm run watch +webserver ... runs the webserver while also building when any file changes (you need to manually refresh)
```

martinheidegger:
> I'd still like to avoid 5 liners but I could probably reassemble them myself afterwards. Or do you see any issues arising if I just merge some files together?

I would differentiate between two case:

1. There is a file that is only used by one other file and it is "inlined".
2. There is a file that is used by more than one other file.

**If you differentiate like this** in the first case, merging the two together, shouldn't be problematic at all because the actually exported value for the other file should not have changed. I actually prefer such unexposed code because it reduces the cost of maintenance immensely. The problem in case of potree is that most (all?) of the functions have been exposed through a public API and as a result the inlining would need to be officially a breaking change.

**If you don't differentiate like this** then we can skip `1.` and are only leftover with case `2.`. An example for that might be `projectedRadius` and `projectedRadiusOrtho`. `projectedRadius` looks like this:

```javascript
module.exports = (radius, fov, distance, screenHeight) => { // ...
```

and if you put them together they might look like:

```javascript
exports.projectedRadius = (radius, fov, distance, screenHeight) => { //...
exports.projectedRadiusOrtho = function (radius, proj, screenWidth, screenHeight) { //...
```

if you do that, the _tricky_ part is the `const THREE = require('three');` (which is part of `projectedRadiusOrtho`). So in order to keep the knowledge that it is only required by `projectedRadiusOrtho` I would put the blocks in braces:

```javascript
{
  exports.projectedRadius = (radius, fov, distance, screenHeight) => { //...
}
{
  const THREE = require('three');
  exports.projectedRadiusOrtho = function (radius, proj, screenWidth, screenHeight) { //...
}
```

Since on of the harder tasks was to figure out which function uses what dependency _(and I would like to avoid that)_.

It is worth noting that if you change it from a direct function export to an export with two keys (projectedRadius and projectedRadiusOrtho) the dependencies also have to change from:

```javascript
projectedRadius = require('../util/projectedRadius')
```
to

```javascript
projectedRadius = require('../util/projections').projectedRadius
```
which **_is_** perfectly legal, but it makes it harder to track _(by just looking at the require tree)_ to which files become obsolete. i.e. `EarthControls` only uses `projectedRadius` but most other references need both. If for some reason the other references all fall away `projectedRadiusOrtho` suddenly becomes unnecessary _(and can be removed from the code-base)_.

`projectedRadius` is furthermore a nice example, because most of the small files that I have found seem like suboptimal ideas in the first place: `projectedRadius` and `projectedRadiusOrtho` are used almost everywhere in the same fashion of:

```javascript
let pr = 0;
if (this.viewer.scene.cameraMode === CameraMode.PERSPECTIVE) {
  let distance = label.position.distanceTo(camera.position);
  pr = projectedRadius(1, camera.fov * Math.PI / 180, distance, domElement.clientHeight);
} else {
  pr = projectedRadiusOrtho(1, camera.projectionMatrix, domElement.clientWidth, domElement.clientHeight);
}
```

Why is this not part of a over-arching `project(...)` method? It smells of copy & paste.
But also other functions such as ...

- `LRU` _(for which probably better solutions exist on npm)_
- `addCommas`/`removeCommas` _(which's uses are questionable form the UX perspective...)_ 
- `createWorker` _(which I would like to replace with a clearer, better workerpool api)_

... look like they could be improved _(refactored away)_ in coming releases. Which is why I found it a waste of time thinking about if they should be in one file or not.

In other cases such as `TreeType` or `PointAttributeNames` ... Enums essentially ... it is worth arguing that they should remain in their respective files for the readability of the code, as they become part of the API effectively.

martinheidegger:
I have managed to get workers running again. I think its kinda ready for a review _(have I missed something)_?

m-schuetz:
> addCommas/removeCommas (which's uses are questionable form the UX perspective...)

Curious, why do you think they are questionable from the UX perspective? I think large values should have thousands seperators, otherwise their order of magnitude is hard to grasp.

martinheidegger:
i will open (and discuss that) in a separate issue.

m-schuetz:
```
D:\dev\workspaces\potree\mh_browserify>gulp watch
[10:11:06] Using gulpfile D:\dev\workspaces\potree\mh_browserify\gulpfile.js
[10:11:06] Task 'watch' is not in your gulpfile
```

m-schuetz:
The build duration increased from ~1 second to around ~16 seconds. What are the bottlenecks here? Any chance to get this down to a few seconds? 16 seconds seems a bit much to watch and debug.

martinheidegger:
@m-schuetz There are ways to improve this but 16 seconds startup time are not so bad, the continuous builds should be quite quick (<1s). The slow parts are - afaik - the las-zip files. Also the building of source-maps _(something that is missing in the current version)_ takes time _(you can try how much faster it is with "debug: false" on  [gulpfile.js#L42](https://github.com/martinheidegger/potree/blob/90eaed1b7262a6039e72fdb4fc7ee59cf2fa90fe/gulpfile.js#L42))_. There is also the case that you can enable local caching to browserify _(where it stores cache files in the hard-disk, which would only make the first time running the build slower)_ but that is still experimental, and I remember there to be bugs in combination with watching.

martinheidegger:
I renamed `gulp watch` to `gulp watch:*` to be more verbose about the fact that there are various watch targets now.

m-schuetz:
The size of the viewer.html page increased from around 5.7MB to 12.7MB, according to the chrome network tracker, with just the root node of the point clouds loaded. How can we get that back to ~6MB?

martinheidegger:
The source code contains sourcemaps now, they are quite big. In the `.min.js` version they should not be present and the amount of code should be reduced.

m-schuetz:
> The source code contains sourcemaps now, they are quite big. In the .min.js version they should not be present and the amount of code should be reduced.

Ah yes, the min version looks good with 5.4MB transferred for the whole viewer.html.

m-schuetz:
```gulp watch:*``` doesn't work for me. It does some stuff on the console but it never listens on 8080 so I can't open the page in the browser:

<details>
<summary>'gulp watch:*' console output</summary>
<pre>
D:\dev\workspaces\potree\mh_browserify>gulp watch:*
[10:50:39] Using gulpfile D:\dev\workspaces\potree\mh_browserify\gulpfile.js
[10:50:39] Starting 'watch:script:main'...
[10:50:39] Starting 'watch:script:bin'...
[10:50:39] Starting 'watch:script:dem'...
[10:50:39] Starting 'watch:script:gre'...
[10:50:39] Starting 'watch:script:las'...
[10:50:39] Starting 'watch:script:laz'...
[10:50:39] Starting 'copy:resources'...
[10:50:39] Starting 'copy:license'...
[10:50:39] Starting 'copy:extra'...
[10:50:39] Starting 'copy:examples'...
[10:50:39] Starting 'copy:pointclouds'...
[10:50:39] Starting 'copy:jqueryUi'...
[10:50:39] Starting 'copy:jquery'...
[10:50:39] Starting 'copy:perfectScrollbar'...
[10:50:39] Starting 'copy:spectrumColorpicker'...
[10:50:39] Starting 'copy:i18nextClient'...
[10:50:39] Starting 'copy:openlayers'...
[10:50:39] Finished 'copy:license' after 297 ms
[10:50:39] Starting 'watch:copy:license'...
[10:50:39] Finished 'watch:copy:license' after 6.23 ms
[10:50:39] Finished 'copy:extra' after 385 ms
[10:50:39] Starting 'watch:copy:extra'...
[10:50:39] Finished 'watch:copy:extra' after 2.12 ms
[10:50:39] 4032 bytes written (0.47 seconds)
[10:50:39] 19131 bytes written (0.49 seconds)
[10:50:39] Finished writing D:\dev\workspaces\potree\mh_browserify\build\potree\workers\DEMWorker.js
[10:50:39] Finished 'watch:script:dem' after 549 ms
[10:50:40] Finished writing D:\dev\workspaces\potree\mh_browserify\build\potree\workers\LASDecoderWorker.js
[10:50:40] Finished 'watch:script:las' after 552 ms
[10:50:40] Finished 'copy:jquery' after 811 ms
[10:50:40] Starting 'watch:copy:jquery'...
[10:50:40] Finished 'watch:copy:jquery' after 3.33 ms
[10:50:44] 33670 bytes written (4.99 seconds)
[10:50:44] Finished writing D:\dev\workspaces\potree\mh_browserify\build\potree\workers\BinaryDecoderWorker.js
[10:50:44] Finished 'watch:script:bin' after 5.13 s
[10:50:48] Finished 'copy:jqueryUi' after 8.75 s
[10:50:48] Starting 'watch:copy:jqueryUi'...
[10:50:48] Finished 'watch:copy:jqueryUi' after 7.94 ms
[10:50:51] Finished 'copy:perfectScrollbar' after 12 s
[10:50:51] Starting 'watch:copy:perfectScrollbar'...
[10:50:51] Finished 'watch:copy:perfectScrollbar' after 15 ms
[10:50:51] 3411337 bytes written (12.28 seconds)
[10:50:52] 3389437 bytes written (12.69 seconds)
[10:50:52] 10188089 bytes written (13.00 seconds)
[10:50:52] Finished writing D:\dev\workspaces\potree\mh_browserify\build\potree\workers\GreyhoundBinaryDecoderWorker.js
[10:50:52] Finished 'watch:script:gre' after 13 s
[10:50:52] Finished writing D:\dev\workspaces\potree\mh_browserify\build\potree\workers\LASLAZWorker.js
[10:50:52] Finished 'watch:script:laz' after 13 s
[10:50:52] Finished 'copy:spectrumColorpicker' after 13 s
[10:50:52] Starting 'watch:copy:spectrumColorpicker'...
[10:50:52] Finished 'watch:copy:spectrumColorpicker' after 34 ms
[10:50:52] Finished 'copy:examples' after 13 s
[10:50:52] Starting 'watch:copy:examples'...
[10:50:52] Finished 'watch:copy:examples' after 50 ms
[10:50:52] Finished writing D:\dev\workspaces\potree\mh_browserify\build\potree\potree.js
[10:50:52] Finished 'watch:script:main' after 13 s
[10:50:52] Starting 'watch:script:*'...
[10:50:52] Finished 'watch:script:*' after 40 μs
[10:50:52] Finished 'copy:resources' after 13 s
[10:50:52] Starting 'watch:copy:resources'...
[10:50:52] Finished 'watch:copy:resources' after 77 ms
[10:50:52] Finished 'copy:i18nextClient' after 13 s
[10:50:52] Starting 'watch:copy:i18nextClient'...
[10:50:52] Finished 'watch:copy:i18nextClient' after 74 ms
[10:50:53] Finished 'copy:pointclouds' after 14 s
[10:50:53] Starting 'watch:copy:pointclouds'...
[10:50:56] Finished 'watch:copy:pointclouds' after 2.88 s
[10:50:57] Finished 'copy:openlayers' after 18 s
[10:50:57] Starting 'watch:copy:openlayers'...
[10:51:04] Finished 'watch:copy:openlayers' after 6.94 s
[10:51:04] Starting 'watch:copy:*'...
[10:51:04] Finished 'watch:copy:*' after 2.47 μs
[10:51:04] Starting 'watch:*'...
[10:51:04] Finished 'watch:*' after 4.23 μs
</pre>
</details>

<br>

Also, when I change something in, for example, src/Potree.js, I'm getting an error:

```
[10:40:38] Finished 'watch:*' after 4.23 μs
D:\dev\workspaces\potree\mh_browserify\gulpfile.js:53
                        .bundle()
                         ^

TypeError: b.bundle is not a function
    at Browserify.bundle (D:\dev\workspaces\potree\mh_browserify\gulpfile.js:53:5)
    at emitOne (events.js:115:13)
    at Browserify.emit (events.js:210:7)
    at Timeout.notify [as _onTimeout] (D:\dev\workspaces\potree\mh_browserify\node_modules\watchify\index.js:149:15)
    at ontimeout (timers.js:488:11)
    at tryOnTimeout (timers.js:323:5)
    at Timer.listOnTimeout (timers.js:283:5)
```

**Comment by martinheidegger** [comment]
Note: It is possible to utilize browserify to get a detailed size profile for all dependencies. In a further step it makes sense to add that. Also: I still believe that the size is quite big. Also there can be other reasons for the increase in size:

- I updated a few dependencies, notable d3.js, and those dependencies could have become bigger.
- browserify *does* have some general overhead (less than 1kb) but also some overhead per file included _(to ensure a consistent API)_, which does make itself noticeable when there are a lot of files required _(which some of the dependencies do)_.
- Some of the dependencies might require some libraries that are never used, those might be present only in the `requirejs` variant of the library (because the library can be used in a server context as well). An analysis of the browserify task could allow us to drop a few.

**Comment by martinheidegger** [comment]
I will take a look at `gulp watch:*` _(will try to deliver an update within 10 hours.)_

**Comment by martinheidegger** [comment]
Note: `gulp watch:*` does not start a webserver. `gulp watch+liveserver` or `gulp watch+webserver` are the tasks (in case you want to run the webserver or liveserver separately for some reason)

**Comment by martinheidegger** [comment]
Pushed commit that should fix `gulp:watch:*`

**Comment by m-schuetz** [comment]
Okay, build performance looks fine while watching, that's good!

It seems like the chrome dev tools are a bit overwhelmed with a single file with 109k lines of code, though. The source tab is unresponsive for half a minute while loading the source. Is there any chance to split the code into potree.js and allTheOtherLibs.js for debugging? Potree itself is only around 20k lines of code as far as I can see.


Also wondering about ```exports.XMLHttpRequest = function() { ...```
XMLHttpRequest is already native in browsers, why is there a module to emulate it?
Are there other modules like that, which could probably be removed?

**Comment by martinheidegger** [comment]
> Is there any chance to split the code into potree.js and allTheOtherLibs.js for debugging? 

It probably is, but I have never attempted it and would need to look into it. I am not sure if it is automatically possible _(though it should be)_.

> The source tab is unresponsive for half a minute while loading the source.

I did experience a similar thing though I wonder if you tried to access each part file (in debug mode):



The "orange" folders are "virtual" folders that point to parts of the potree.js file _(basically what the sourcemap is here for)_. And those should be very quick.

**Comment by m-schuetz** [comment]
> The "orange" folders are "virtual" folders that point to parts of the potree.js file (basically what the sourcemap is here for). And those should be very quick.

You're right, that works much better. No need to split then, I guess.

**Comment by martinheidegger** [comment]
Regarding: `exports.XMLHttpRequest`... before I mentioned that: 
> those might be present only in the requirejs variant

It seems that you found a implementation detail of the new `d3` version: `d3-request`.

**Comment by martinheidegger** [comment]
d3 was simply referenced in the last implementation you might have tested. I saw that d3 actually consists of a part of implementations: [`d3/index.js`](https://github.com/d3/d3/blob/635d44786dd4322d0c89dd98546bff91d7cce0fe/index.js) so I replaced `d3` with the parts that we actually use. This removed the dependency to `d3-request` and also the export of a `XMLHttpRequest`. The resulting .js file should have become noticeably smaller.
Also I updated openlayers to the latest version because it was a rather simple update.

**Comment by m-schuetz** [comment]
Getting an error while trying to select another language in the sidebar:
image

According to the chrome dev tools, ```viewer.setLanguage(``` is being executed, so the argument and closing bracket are missing.

The chrome DOM inspector claims that the html code for this element is ```<a href="javascript:viewer.setLanguage(" fr');"="">FR</a>```
There seems to be a mixup between single and double quotes. In the original source in src/viewer/sidebar.html the js statement is within double-quotes and the argument within single-quotes: ```<a href="javascript:viewer.setLanguage('fr');">FR</a>```

**Comment by martinheidegger** [comment]
@m-schuetz thanks for checking that, found the issue and pushed a fix.

**Comment by m-schuetz** [comment]
Looking good so far. Just las/laz support missing. Anything else?

**Comment by martinheidegger** [comment]
Las/laz is working here, what is broken?

**Comment by m-schuetz** [comment]
Trying to load examples/ca13.html:

```
Uncaught TypeError: Promise.resolve(...).cancellable is not a function
    at LasLazLoader.parse (LasLazLoader.js:66)
    at XMLHttpRequest.xhr.onreadystatechange (LasLazLoader.js:52)
```

**Comment by m-schuetz** [comment]
Hm, it seems like this comes from bluebird.js Promises and as far as I can see only LasLazLoader uses that.
I'll check if we can just remove bluebird and use ES6 Promises instead.

**Comment by bbaia** [comment]
Use ES6 promises instead with a polyfill:
https://github.com/stefanpenner/es6-promise
es6-promise-auto-min 6.19 KB (2.4 KB gzipped) - Minified version of es6-promise-auto above.

**Comment by m-schuetz** [comment]
@martinheidegger Can you add me as a collaborator to your browserify fork? I'd like to commit some changes that remove bluebird.

**Comment by m-schuetz** [comment]
Ok, merged to potree:browserify now. Will do some more checks and changes here but things are looking good now. Thanks for that PR!

**Comment by martinheidegger** [comment]
I added <a href="https://gyazo.com/533bc5e058062d8bd651977232e05f65">https://gyazo.com/533bc5e058062d8bd651977232e05f65</a>
which should allow you to edit the branch directly. _(It was always possible to send PR's to the fork branch!)_. I don't understand why you would mark it as merged if its not yet merged?!

**Comment by m-schuetz** [comment]
It was automatically marked as merged by github since I merged it into potree:browserify, which I'm going to merge into potree:develop after some further checks.

**Comment by m-schuetz** [comment]
Merged into develop now. Thanks again for that PR!

**Comment by paubineau** [comment]
Hey, quiet new with potree here. 
I was trying to use this branch in order to integrate potree in a react-based application built with webpack. It works quiet well, but I had an error while loading a pointcloud using viewer.scene.addPointCloud(e.pointcloud) : 

```
Uncaught Error: override function
    at PointCloudOctreeNode.isTreeNode
```
Looking for a fix, I found that removing that line :
`PointCloudOctreeNode.prototype = Object.create(PointCloudTreeNode.prototype);`
from src/tree/PointCloudOctreeNode.js (Same way as it is written down as a TODO in PointCloudArena4DNode.js)  allows me to load my model.

I do understand both class should inherit from PointCloudTreeNode and throw an error if some specific method are not implemented, but it wasn't working as it is on the branch right now. Is it normal ? Did I make something wrong ?

**Comment by martinheidegger** [comment]
@paubineau I think that in the PR I introduce those error messages. For consistency both should implement it. It would be super-helpful if you could submit a PR that implements it as a stub 

```javascript
isTreeNode = function() {
   console.warn('Only stub implementation for isTreeNode, a PR that implements this properly would help a lot!')
   return false
}
```
in `PointCloudOctreeNode` & `PointCloudArena4DNode`.

adrelino:
It seems like this PR is still not merged into the branch `develop`, but rather into `develop_2017.10.10`. Also, the source-branch `marthinheidegger:browserify` is deleted. (When) @m-schuetz do you plan to merge into `develop` and use the huge changes this PR introduced as a base for all future work?

m-schuetz:
@adrelino While it's nice to have a browserify build system, it introduces some issues that made debugging and profiling very difficult in chrome, so I reverted it for now in order to be able to continue work. I'd like to check browserify again in the future but probably not for the comming release.

PR #399, opened 2017-09-06
Removing `-` from beginning of each line in LICENSE

corymsmith:


martinheidegger:
Woops, I mistook them for Diff lines in the changes view 😅

PR #412, opened 2017-09-22
Removed /build folder as it is part of .gitignore

martinheidegger:
In the [`.gitignore`](https://github.com/potree/potree/blob/cb25df30b9a10a053c5f97d0d93f71bd51c44474/.gitignore#L1) file `/build` is marked as "should not be commited". But since at some point they got commited, they will be marked as changed and still be contained repository. This PR removes the build folder entirely to avoid unnecessary merge conflicts.


m-schuetz:
Hi Martin, I did that so that people won't submit their build folders. I still want to comit it myself, though.

martinheidegger:
Running a build on my computer will still show the changes to the current files in the build folder as "commitable", with the only effect that new files (new resources etc.) would not be shown commitable. Effectively rendering the current `/build` entry in the `.gitignore` more harmful than helpful.

It isn't all clear to me: why keep the build folder in the repository? I see no practical value?!

m-schuetz:
I guess you're right, should be sufficient to build for release.

martinheidegger:
I am sorry, but it isn't quite clear: Do you want me to merge the change? Am I supposed to wait for you to merge it?

m-schuetz:
Sorry, I was writing from my mobile phone which is why I haven't merged it. Doing that now. :)

Issue #424, opened 2017-10-05
Performance problems with unminified Potree


m-schuetz:
Potree now takes around 6 seconds to load for the unminified code but

martinheidegger:
The reason why it is slow is likely the loading of Source Maps. I remember mentioning it in #391 but Source Maps are good for us because people can track errors to actual files in the repository, which should lead to better error reports.

That aside: I am working on a PR that splits the sourcemap from the .js files; which should make it easier for you to see the differences.

Side question: Did you try it in the recent Firefox 57? _(it has become a lot faster)_

m-schuetz:
After the PR, debugging is a bit broken.
When I set a breakpoint somewhere, chrome now has difficulties resolving the correct variable values,
More often than not, chrome claims variables are undefined on a breakpoint even though they are not.
Sometimes, chrome now also confuses one variable name with another and I can't set breakpoints on many lines.

m-schuetz:
build/potree/potree.js now contains the minified source. Can I have the unminified source again?

PR #427, opened 2017-10-05
Split source map

martinheidegger:
Source maps can take quite a while to load as mentioned in #424. Also: It is possible that someone wants to debug the minified source code (which takes quite a bit longer to build). So this PR is more of an idea how to do it differently: It will simply generate source maps in both cases and puts them in the same directory as the built files.

_Note: This PR assumes that #426 will be merged before this._


m-schuetz:
Thanks!

PR #436, opened 2017-10-25
fix typo in package.json

adnan-wahab:
this causes any project installing potree via commit hash via yarn to throw an error


m-schuetz:
Thanks for the PR! It seems like it is no longer necessary, though, since package.json doesn't contain these entries anymore. Closing this.

PR #439, opened 2017-10-31
simplify worker setup

calvinmetcalf:
This pull adds some abstraction around the workers which should simplify using them considerably

1. brings them together as a single file instead of 4 separate workers, this should make queuing simpler and also makes the build simpler.
2. put some abstractions around the worker pool to prevent some theoretical memory leaks and to catch some errors
3. generally make it easier to add more workers in the future.

this doesn't add middleware to fold the workers into a single file because `workerify` doesn't support es6 due to an outdated dep and webworkerify doesn't actually seem to do what we want I don't think.


package.json
```diff
@@ -61,7 +49,10 @@
     "release": "npm run gulp -- release"
   },
   "main": "./src/index.js",
-  "browser": "./src/index.js",
```
martinheidegger: Thanks, that would have been an awesome PR ❤️

gulpfile.js
```diff
@@ -20,11 +20,7 @@ const through = require('through');
 
 	const SCRIPTS = {
 		main: {source: 'src/index.js', target: 'build/potree/potree.js', args: {standalone: 'Potree'}},
-		bin: {source: 'src/workers/BinaryDecoderWorker.js', target: 'build/potree/workers/BinaryDecoderWorker.js'},
-		dem: {source: 'src/workers/DEMWorker.js', target: 'build/potree/workers/DEMWorker.js'},
-		gre: {source: 'src/workers/GreyhoundBinaryDecoderWorker.js', target: 'build/potree/workers/GreyhoundBinaryDecoderWorker.js'},
-		las: {source: 'src/workers/LASDecoderWorker.js', target: 'build/potree/workers/LASDecoderWorker.js'},
```
martinheidegger: The LasDecoder is very big (1MB or so), would be good if - at least - that one would stay its own worker.

src/WorkerPool.js
```diff
@@ -1,25 +1,46 @@
...
-		let worker = this.workers[url].pop();
+		let worker = this.workers.pop();
 
 		return worker;
 	}
 
-	returnWorker (url, worker) {
-		this.workers[url].push(worker);
+	returnWorker (worker) {
+		if (this.workers.length >= this.max) {
```
martinheidegger: I think this would be better implemented in a different way? It might be hard to debug it if the worker is accidentally added to the wrong worker list.

```javascript
worker.returnToWorkers = function () {
   // Return it to the list
}
```
might be better?

package.json
calvinmetcalf: yeah I can break this out, I noticed that crypto was showing up in the build so this prevents it

gulpfile.js
calvinmetcalf: so that actually might be because of crypto erroneously getting bundled into that one, let me pull that part out

src/WorkerPool.js
calvinmetcalf: the key here is just that since there is only a single list of workers, so it would never get added to the wrong one

gulpfile.js
```diff
@@ -20,11 +20,7 @@ const through = require('through');
...
-		laz: {source: 'libs/plasio/workers/laz-loader-worker.js', target: 'build/potree/workers/LASLAZWorker.js'}
```
calvinmetcalf: woops we forgot about this one

gulpfile.js
calvinmetcalf: the las decoder worker is only 2.5k its the greyhound and laz loader that are large, ~700k with the crypto fix, but the actually are relying on the same library so combining those 2 would save the user from having to download laz-perf twice, though I realized I didn't include laz loader in this pull

calvinmetcalf:
ok I've got all 5 workers into a single file clocking in at 17k minified by having the 2 scripts that need laz-perf lazy load it the first time they need it.

src/WorkerPool.js
martinheidegger: If there is only one "list" then the url shouldn't be part of the constructor, no?

martinheidegger:
Linting ;)

src/WorkerPool.js
calvinmetcalf: so there is only one list per worker pool, but at some latter date we might want to have multiple worker pools so having it as a parameter makes it more flexible in the future, also we could hard code it but it doesn't have access to context so we'd probably have to pass that in anyway.

calvinmetcalf:
ok fixed linting

PR #451, opened 2017-12-04
Oneline Hotfix for cloud transparency

jo-chemla:
When transparency was set lower to one, `this.depthFunc = THREE.AlwaysDepth`, but it was then not set back to `THREE.LessEqualDepth` when cloud opacity/alpha was equal to 1 again.


m-schuetz:
Thanks, @jo-chemla 

It's probably only doing something for point clouds in the profile window, since point clouds in the 3d window are not rendered with three.js anymore. I'll work on fixing that.

Issue #449, opened 2017-12-03
Potree 1.6 issues and todos


m-schuetz:
Collection of smaller issues to be fixed for 1.6

- [ ] lion.html and multiple_pointclouds.html: lion dataset being funny

<img src="https://user-images.githubusercontent.com/6705073/33525142-e206c3a4-d829-11e7-95c0-222cdd1a7c51.png" height="200px"> <img src="https://user-images.githubusercontent.com/6705073/33525201-6baeeb3a-d82b-11e7-983e-ed9989fcf539.png" height="200px">

- [ ] arena4d.html not working
- [ ] entwine_greyhound.html not working
- [ ] [cesium examples only] double click away from the point cloud zooms somewhere into the point cloud. 
- [ ] Invalid extents in profile window, usually reproducible in ca13.html with one marker at the border of the point cloud.
- [ ] In lion_las.html and laz point cloud doesn't show up

m-schuetz:
# Done
Verify again before release

Collection of smaller issues to be fixed for 1.6



### Demos 

- [x] lion.html and multiple_pointclouds.html: lion dataset being funny [Fig. 1](https://user-images.githubusercontent.com/6705073/33525142-e206c3a4-d829-11e7-95c0-222cdd1a7c51.png) [Fig. 2](https://user-images.githubusercontent.com/6705073/33525201-6baeeb3a-d82b-11e7-983e-ed9989fcf539.png)
- [x] arena4d.html not working
- [x] entwine_greyhound.html not working
- [x] [cesium examples only] double click away from the point cloud zooms somewhere into the point cloud. 
- [x] In lion_las.html and laz point cloud doesn't show up
- [x] Create a demo with a textured mesh. The PLYLoader only supports untextured meshes.
- [x] Create a demo for hierachical annotations.



### Profile Window

- [x] Profile Window unnecessarely slow during drag of profile markers.
The picking itself is just around 3-6ms, the redraw of the Profile Window and the Panel in the sidebar account for most of the time with another ~25ms. This makes the profile tool feel very slow and sluggish while a marker is dragged. [Fig. 1](https://user-images.githubusercontent.com/6705073/33659157-7e7084da-da7f-11e7-8a09-4443ebcf000a.png)
- [x] Invalid extents in profile window, usually reproducible in ca13.html with one marker at the border of the point cloud. (#452)
- [x] Profile Window quite slow while points are still being loaded. 
- [x] Profile Window jumping around while zooming in
[...]

PR #463, opened 2017-12-28
Merging develop into develop_2017.10.10 (WIP?)

martinheidegger:
In this PR I merged all the changes that have been done on the develop branch to the develop_2017.10.10 branch. Since I tried to maintain the linting style, it has been a **lot** of work getting this PR to the current state. I have not had the patience yet to test this thing, which is why i put WIP in the title. I am pretty sure there are some bugs in the code but those should "just" be the same bugs as in develop (while merging I noticed some code that can't work). I plan to further test this before merging, but if anyone feels like lending me a hand, that would be very warmly welcome.

@m-schuetz I hope you can give a browserified version another chance?


m-schuetz:
I'll give it a look after releasing 1.6 (in January).
One of the biggest issues last time was, as far as I remember, that debugging in chrome didn't work properly anymore.

DanielSmon:
Hey guys, we (Synth) are back on to this now after a few other priorities cropped up over the last few months. Keen to help with any blockers such as the dev / debugging experience as this will soon become quite important for us as well.

martinheidegger:
@DanielSmon I saw that @m-schuetz added another few dozen of commits to develop which needs to be merged in this branch. I am a bit busy with other work right now to focus on it. It is not easy work to continue merging as I did (manual labour + understanding of code) but if you want to give it a try with one or two commits: I would love to see a PR to this PR.

DanielSmon:
Hey @martinheidegger, I've had a good look at this and you're right that it's rather complicated :). It may be a little beyond me unfortunately (within a reasonable time frame at least).

I just want to make sure I'm clear on the delta though. If I do a compare between *martinheidegger:develop_2017.10.10_update* and *potree:develop*, is that an accurate picture of the commits to be applied to *martinheidegger:develop_2017.10.10_update*?
image

Or is it more accurate to manually look through the commits on develop from 14 Dec onwards and manually apply the relevant changes, i.e. without a merge? Thanks for clarifying :).

m-schuetz:
I'll have to say, I'm still not convinced that the effort of browserifying Potree outweights the benefits.
Last time, I had serious troubles debugging in chrome afterwards and building took about two orders of magnitude longer. Right now, I'm very happy to have "gulp watch" instantly create a new build on pressing "ctrl+s" with next to no startup time at the first time. 

Btw, what about ES modules? Wouldn't it make sense to wait for them to take off, now that chrome, firefox and Safari have basic support?

martinheidegger:
The reason I didn't use es6 modules instead of require was to have LESS changes in one PR. ES6 modules would increase the amount of lines even more & in order to have a build that works now it would be even slower than browserify. That you had trouble debugging in chrome was partly depending on bugs in chrome with sourcemaps. Just because it supports ES6 doesn't mean it implements es6 files without bugs (i.e. wrong files to find). Also: if es6 modules are supported in the browser: all the modules would need to be exposed to the browser (which would put a penalty on modularization). In the current build browserify also packs webworker into modules. Something that would require a different solution.

DanielSmon:
Hey @m-schuetz, in response to why we think this is useful there are two main reasons from our perspective. We are effectively consuming potree, integrating it in to (what will become) a much larger application. Without going in to detail, having modularised dependencies makes them much easier to integrate in to our development process. This includes applying consistent optimisation like minification etc.

Secondly, as someone who still isn't that familiar with the code-base, navigating and reading the commonjs version of the project is much easier. Especially as my editor (VSCode), as well as many others can immediately jump to code definitions. Comparing this with the current project structure, I have to use Ctrl+F and manual navigation to poke around. This is also relevant to your comment regarding having so many files. For me this is simply a non-issue as my editor makes this easy to manage. If anything, for me having more files is easier as it is simpler and quicker to switch between files and understand their contents as opposed to remembering exact scroll positions or line numbers to jump between. I admit, this is partially subjective though :).

So for us, we're considering both the development "inner loop" (e.g. ctrl+s, F5) for potree, as well as the development inner loop for our application that consumes it. I am confident that we can get the developer inner loop for potree optimised so that we can have the best of both worlds! :).

martinheidegger:
@DanielSmon Because there have been many changes in `develop_2017.10.10`-branch I went and looked at every commit: checked the changes and replicated them as commit in `develop_2017.10.10` branch.

Lets look at this with one example. Take this commit: https://github.com/potree/potree/pull/463/commits/0fbb9bb2cf5f93341492508546d68de6f56d7f37,  the commit-messages says:

> "fixing some smaller issues #449 (ref: [#54e86cc](https://github.com/potree/potree/commit/54e86ccf4f862955f140c2654b396f58a288e01b))"

In the end it stats `ref #54e86cc` by which I mean: 

> The changes that also occur on the develop branch under commit #54e86cc

So what I did there was look what changes happened in #54e86cc and then replicated them in the 
 update branch; ran the linting tests against them _(only to see if I made the situation worse, linting is broken right now)_ and used the same commit message + ref for the commit https://github.com/potree/potree/pull/463/commits/0fbb9bb2cf5f93341492508546d68de6f56d7f37 - does this make sense?

m-schuetz:
> only to see if I made the situation worse, linting is broken right now

I'll handle the linting when I've got some time. I want to do some changes to the config.

Is it somehow possible to do a softer transition to modules, e.g. start to turn a few of the easier classes into modules and then add more over time?

martinheidegger:
@m-schuetz This is really as soft as I was able to imagine & deal with. There are always other ways to do things though.

DanielSmon:
Cool, thanks @martinheidegger. That makes sense now. I will take a closer look over the coming days.

PR #472, opened 2018-01-17
Implementation of factory to supply XHR, allowing customisation of XHR requests

### LINKED ISSUE #408: Add support to send credentials with xhr requests to render content that is protected
Currently there is no way to send credentials like cookies and authorization headers with the xhr requests.

Adding support for sending credentials when making xhr requests will enable potree library to render `content that is protected/secured`.

I see calls like:
```
		let xhr = new XMLHttpRequest();
		xhr.open('GET', url, true);
```

Changing xhr calls by setting `withCredentials` to true should send the credentials to the secured CDN that is hosting the potree data files
```
		let xhr = new XMLHttpRequest();
		xhr.withCredentials = true;
		xhr.open('GET', url, true);
```

I understand that the code needs to be refactored better to accomplish this. Is the potree team willing to entertain a PR to `Add support for sending credentials`? I will work on the PR.

DanielSmon:
Hi Markus,

For your consideration, this PR swaps out instances of "new XMLHttpRequest" with an XHRFactory. It's a bit of a stop-gap measure that we need to move forward. I don't mean to distract from the commonjs and other efforts! On the topic of implementation, I like @martinheidegger 's suggestion to use a template for the fetch API, but that would require a little bit more work. Plus I already had this one implemented as a prototype so thought it would be a quick win which might be helpful for other users as well.

The default implementation of the XHRFactory allows some header customisation, alternatively this allows consumers to override the XHR factory for their own custom implementation for ultimate flexibility. I actually prototyped this a while ago to allow customisation of XHR requests with custom header / auth information but only now got around to cleaning it up for a PR.

This is also effectively resolves #408, although with a slightly different approach than the one described in that issue.

The only other change you'll notice is to force *files.trimTrailingWhitespace* to *false* in VSCode settings. I did this because I have that setting set to *true* locally which created a messier merge. This way, other VSCode users will have an easier time creating PRs if they have that setting set to *true* like me :).

Cheers,
Daniel


m-schuetz:
Looks fine, thanks!

Issue #394, opened 2017-08-31
Potree Annotations Clear


michelleyao1104:
Hi everyone,

I'm new to use Potree 1.5 version. and i find a cool item, is annotations.
Like this url ([example] (http://potree.org/demo/potree_1.5/examples/annotations.html))

I try to clear all annotations, and load other new annotations to scene,
but i did not find clear annotations function in potree project,

I want to know clear annotations features can do it?
So can you tell me how to clear annotations when button click or function?
Thank you!

martinheidegger:
Hello @michelleyao1104,

I tried to do clear the annotations like this:

```javascript
let annotations = viewer.scene.getAnnotations();
annotations.children.forEach(anno => {
	annotations.remove(anno);
	anno.dispose();
});
```

and indeed the annotations were not cleared properly. Since there is a "remove" method and a "dispose" method. It seems like this should work. Which is why this, not working, seems to be a bug.

cherukumilli:
@martinheidegger - I am running into the same issue.
Were you able to find a workaround for this bug yet? If so, can you please post it here?

koala365:
please try this:

var annotations = window.viewer.scene.getAnnotations();
		annotations.children.forEach(function(anno) {
			window.viewer.renderArea.removeChild(anno.domElement.get(0));
			annotations.remove(anno);
			anno.dispose();
		});

reshu-vaish:
@koala365 

While using this snippet 
var annotations = window.viewer.scene.getAnnotations();
annotations.children.forEach(function(anno) {
window.viewer.renderArea.removeChild(anno.domElement.get(0));
annotations.remove(anno);
anno.dispose();
});

Annotation not removing from sidebar which is displaying inside the annotations block. 

Any idea ?

PR #487, opened 2018-03-05
Allows removal of an annotation

rob-waring:
WRT https://github.com/potree/potree/issues/394

This PR allows removal of an annotation via the Scene. It also removes all the children of that annotation when it's removed.

Thanks

Rob

### DIFF
--- a/src/Annotation.js
+++ b/src/Annotation.js
@@ -206,9 +206,27 @@ Potree.Annotation = class extends THREE.EventDispatcher {
 		}
 	}
 
+	hasChild(annotation) {
+		return this.children.includes(annotation);
+	}
+
 	remove (annotation) {
-		this.children = this.children.filter(e => e !== annotation);
-		annotation.parent = null;
+		if (this.hasChild(annotation)) {
+			annotation.removeAllChildren();
+			annotation.dispose();
+			this.children = this.children.filter(e => e !== annotation);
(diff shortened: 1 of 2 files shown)


src/Annotation.js
```diff
@@ -206,9 +206,27 @@ Potree.Annotation = class extends THREE.EventDispatcher {
...
 	}
 
+	hasChild(annotation) {
+		return this.children.includes(annotation);
+	}
+
 	remove (annotation) {
-		this.children = this.children.filter(e => e !== annotation);
-		annotation.parent = null;
+		if (this.hasChild(annotation)) {
```
m-schuetz: Won't this always evaluate to false, since you've removed the child from this.children two lines before?

src/viewer/Scene.js
```diff
@@ -5,7 +5,7 @@ Potree.Scene = class extends THREE.EventDispatcher{
 		super();
 
 		this.annotations = new Potree.Annotation();
-		
+
```
m-schuetz: No changes to formatting in the PR, please

m-schuetz:
Thanks for the PR!

Gwhitch:
How is this accomplished? All examples I have run across do not remove the annotations. Do you have a small example of the script calls I need to make? Is there another script I have to load?

PR #495, opened 2018-03-17
3d shapefile

paulopires16:
Is it possible to view 3d shapefile with your Potree?
Your shapefile example is 2D... converting to json for 3D?
json.html file as an example...


m-schuetz:
I'm very confused by this PR, and why it shows up as merged. I'm not aware of doing this. o_O

PR #499, opened 2018-03-23
Fix a Stats() undefined error at viewer.html

wemamawe:
Fix a Stats() undefined error  at viewer.html and should add a potree.css to Build folder


potree:
Hi @wemamawe 

I think there may be a missunderstanding. 
- Potree doesn't use Stats.js anymore. Are you sure you've updated your fork to the latest version of the develop branch?
- The build folder is no longer comitted. When cloning the repository, you get the source code only and have to do the build yourself.

wemamawe:
yes,  i download 1.6zip file from github.com again, and still has this problem.

m-schuetz:
I just checked the Potree_1.6.zip file and I can't find any occurance of Stats being in use.

PR #508, opened 2018-04-20
xhr request support loading from file system

ozvale:
when i start chrome with --allow-file-access-from-files,
it will return status 0 when i start a XMLHttpRequest for las/laz file.


m-schuetz:
thanks

PR #555, opened 2018-09-19
Turn on/off visibility of children nodes in sidebar

### LINKED ISSUE #513: Toggle group on Scene Sidebar doesn't work
Using the [Annotations example](http://www.potree.org/potree/examples/annotations.html) I am unable to toggle the visibility of a group from the scene sidebar, but I can toggle individual items. Here is a demo:
potree_group_select

I see this in the latest Firefox and Chrome running on Ubuntu 16.04

### LINKED ISSUE #521: Fix "three_state" behaviour when set to "true"
I thought it would be great to get that thing going, so that it would be possible to turn off/on the visibility of whole nodes.
I've turned the "three_state" to "true" in potree.js but it doesn't work. 
Could you give me some hints where would be the best way to fixing this? I'm new to JS, but I would like to contribute. Any hint would be appreciated :) Below are screenshots of the situation happening. (checkboxes are unchecked, but models are still visible).
image
image

### LINKED ISSUE #534: Sidebar Visibility Toggle
Hi,

Long time reader of this forum, first time poster. I was curious if anybody found a solution for the sidebar group object visibility toggles. The individual object visibility toggle functions properly, but the group visibility toggle does not. For instance, I have several annotations in my scene that I would like to toggle off all at once, but I can only toggle them off one by one. Any help or insight would be more than appreciated! 

image

gpbmike:
This PR changes the behavior of jsTree within the sidebar.

The `checkbox.three_state: false` option has been removed in favor of the default `true` value so cascading works. https://www.jstree.com/api/#/?f=$.jstree.defaults.checkbox.three_state

Logic to handle children of the checked/unchecked nodes has been added. Child nodes of the checked/unchecked nodes will have their visibility toggled.

fixes #513 
fixes #521 
fixes #534 

sidebar toggles

### DIFF
--- a/src/viewer/sidebar.js
+++ b/src/viewer/sidebar.js
@@ -293,7 +293,6 @@ export class Sidebar{
 			},
 			"checkbox" : {
 				"keep_selected_style": true,
-				"three_state": false,
 				"whole_node": false,
 				"tie_selection": false,
 			},
@@ -436,6 +435,13 @@ export class Sidebar{
 			if(object){
 				object.visible = false;
 			}
+			
+			for (let i = 0; i < data.node.children.length; i += 1) {
+				const node = tree.jstree('get_node', data.node.children[i])
+				if (node.data) {
+					node.data.visible = false;
+				}
+			}
 		});
 
 		tree.on("check_node.jstree", (e, data) => {
@@ -443,6 +449,13 @@ export class Sidebar{
 
 			if(object){
 				object.visible = true;
+			}
+			
+			for (let i = 0; i < data.node.children.length; i += 1) {
+				const node = tree.jstree('get_node', data.node.children[i])
+				if (node.data) {
+					node.data.visible = true;
+				}
 			}
 		});
 


m-schuetz:
Nice, looks useful. Love the gifs to show what it does.

It conflicts a bit with the way hierarchical annotations are currently handled, though. They get special treatment (see examples/testing/annotations_performance.html) but maybe I'll just remove that special treatment and make them work the same way as pointclouds. Or do you know if it's possible to keep checkbox.three_state = false within the annotation subtree?

jrobichaud:
This fixes a big usability problem. Please merge.

WargodHernandez:
I implemented a similar feature prior to seeing the issues/PRs
capture-1
I didn't use three_state instead just looping through (parents & children_d nodes) to cascade, the benefit of this method is that it preserves the child visibility state. for my use case it makes more sense then having the parent wipe away the child state.

Maybe we can combine the methodologies somehow and have the three state of the parent be (All off, Previous State, All on). I think that would an extra bool per tree node and the supporting logic, but would seem like the best option.

mccarthyryanc:
Please merge, this fix would really help me out.

mccarthyryanc:
@m-schuetz any chance that the suggestion from @WargodHernandez can be applied? Would really like to have this feature!

Tofoillote:
Hi ^^

mmh sorry I have to ask, but, the one I see in Gifs is exactly how I think it is supposed to work, but i'ts not.

does this depend of "outside" potree code dependancy ?

PR #563, opened 2018-10-09
Separate measure display units from measure data units

rburgstaler:
Add the ability to have the measure display units be different than the units used in the point cloud

The setLengthUnit() method that existed before only set the labels that display when measurements are being made.  This is great when the data happens to be in the same units that the end user thinks in.  To separate the concerns of data vs. user interface a new method setLengthUnitAndDisplayUnit(lengthUnitValue, lengthUnitDisplayValue) was created which takes two parameters: lengthUnitValue which indicates which units the data is in and lengthUnitDisplayValue which indicates which units the end user would like to see the measurements displayed in.  The legacy setLengthUnit() will still operate as it always has.

As an example, one can set the units to be display as feet when the data is in meters with the following call:  `viewer.setLengthUnitAndDisplayUnit('m', 'ft');`

### DIFF
--- a/src/utils/Measure.js
+++ b/src/utils/Measure.js
@@ -373,7 +373,8 @@ export class Measure extends THREE.Object3D {
 				let distance = point.position.distanceTo(nextPoint.position);
 
 				edgeLabel.position.copy(center);
-				edgeLabel.setText(Utils.addCommas(distance.toFixed(2)) + ' ' + this.lengthUnit.code);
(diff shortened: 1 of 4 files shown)


gpbmike:
I had the same requirement and approached it in a different way. Curious what you think.

https://github.com/3drobotics/potree/commit/1c4d574ff7d9c602de476aadf1b0c5ba3b69c30a

If `formatLabel` is set on the measurement, it will be used to format all labels, leaving it up to you to return a string back to potree to render.

m-schuetz:
Looks good, thanks for the PR!

PR #567, opened 2018-10-28
Update potree-file-format links

JeGoi:
It looks that liblas.org as moved from https://www.liblas.org to https://liblas.org without redirection
Same for laszip.org


m-schuetz:
Thanks for the fixes!

PR #573, opened 2018-11-12
Support Matcap for Potree

jo-chemla:
When normals are available per point, can support matcap by computing reflected eye vector at each location, and infering color from a matcap texture. 30 matcap textures included, from the web.

koc_matcap_2
koc_matcap_3


m-schuetz:
Just took a quick look, that looks great for point cloud with normals. 
Some changes will be necessary so that it doesn't affect point clouds that don't make use of this feature though. For example, pointcloud.vs declares a sampler2D for matcap, even if matcap isn't used. With WebGL1 or on lower end GPUs, uniform and sampler spots are very limited and you may end up with shader compilation errors.

I'll have a closer look in the next few days.

m-schuetz:
TODOs:

[ ] Missing sidebar support. Currently set with ```viewer.scene.pointclouds[0].material.pointColorType = Potree.PointColorType.MATCAP``` but this causes a sidebar related error message in the dev console.
[ ] In the vertex shader, encapsulate matcap related attributes and functions (matcapTextureUniform, getNormal, getMatcap) within ifdefs to declare them only when they're used.
[ ] matcap demo in examples.html (demos also serve as tests)

Nice to have: 
[ ] Option to select matcap texture in the sidebar.

jo-chemla:
Hi markus, thanks for the reply and insight. I've updated the PR with the fixes you mentioned 
 - an example, with lion-takanawa with normals as potree, 
 - a way to select matcap texture in UI, 
 - encapsulated  matcap related shader code by ifdefs. 

Here is a preview. Please let me know if there are things I should do differently. 
image

m-schuetz:
I love this! I'll merge it the next days.

m-schuetz:
TODOs (for me): 

- [ ] Add note where the matcap files are from. ("/Blender/2.80/datafiles/studiolights/matcap/*")

m-schuetz:
Merged. Thanks for that great PR!

PR #578, opened 2018-11-26
set leaf of tree band to parent 

wemamawe:
i add some code to sidebar.js, it can tie the visible of object with its parent branch.
i have test it,and have a good effect.


m-schuetz:
Hi, thanks for the PR. 
Could it be that this is a duplicate of this earlier PR?
https://github.com/potree/potree/pull/555

wemamawe:
hi , i read code of #555 and compare it with mine, i think its had the same effect, but I consider more situation, and write some edge condition. so , you can think about my code and make your decision.

wemamawe:
Another thing is about memory leak, I found when I view a huge Las file,like more than 10GB, The Potree always be crash, bucause of out of memory, then I disable the sanbox of Google Chrome, and use DVT to analysis in timeline, the memory grow more then 4G, it is unacceptable.
 So I go to look the PR code deeply, The PR use Potree.Renderer to render the pointcloud, and use a buffers to store geometry that had been loaded. then the buffers grow all the time until out of memory.
 I limit the size of buffers to 1000, and use a LRU function to decide which geometry will be store, the fix this problem.

wemamawe:
add chinese translation

PR #587, opened 2018-12-14
EPT: rename ticks->span

connormanning:
See connormanning/entwine#151.  EPT 1.0 will be released in a few days - this should be the last format-related change before it's cemented.

### DIFF
--- a/pointclouds/lion_takanawa_ept_bin/ept.json
+++ b/pointclouds/lion_takanawa_ept_bin/ept.json
@@ -1,5 +1,5 @@
 {
-	"bounds" : 
+	"bounds" :
 	[
 		-7,
 		-1,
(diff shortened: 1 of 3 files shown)


connormanning:
Since the format was in a dev version prior to the 1.0 release of the EPT format a couple of days ago, I think the number of people that might be affected by this change are small.  I've just committed a fallback so that previous data may still work properly.  Prior to the release of EPT, it was made clear to the early adopters that pre-1.0 data was just that: early adoption of a dev format.  This is no longer true since we've finally released.

There is a required key in the EPT metadata called `version` which is currently `"1.0.0"`, so going forward now that the initial format is specified, things like this will be versioned and handled properly.

potree:
Great, thanks!

Issue #606, opened 2019-02-07
Fix elevation slider range clamping bug


rjpals:
Initially reported here: hobu/usgs-lidar#2
Using elevation range slider causes bounds to default to 0-1000, regardless of the actual range of the data.

sunapi386:
I just saw the same issue, was going to file a ticket as well. 

![out](https://user-images.githubusercontent.com/1029022/52456200-8ace0100-2b08-11e9-91f8-991d231a7c35.gif)

PR #607, opened 2019-02-07
Fix typo. Closes potree/potree#606

rjpals:
Fix typo that leads to range clamping bug as described in potree/potree#606 and hobu/usgs-lidar#2


potree:
Great, thanks!

PR #612, opened 2019-02-17
EPT profile viewing.  See connormanning/potree#2.

connormanning:
Fixed profile viewing for EPT data.  Adds fallback logic for usage of `hierarchyStepSize`, which doesn't exist in the EPT format.


potree:
Thanks!

PR #635, opened 2019-04-06
Solve memory leak on point attributes in PointCloudOctree.js

HeadLessHUN:
I found one of the memory leaks in potree. It is mainly because the attributes of the removed point's kept in memory. Properly delete/dispose these attributes before geometry gets disposed.

To produce memory leak:
-open an example
-open dev tools
-create memory snapshot look for the size
-set pointbudget to minimum and then maximum 
-create memory snapshot and the sizes always gets higher...

After this commit it will stay similar no matter how many point budget settings have done. 

I don't have the possibility to create this for Potree.PointCloudArena4D's disposeListener. But i think it is required there too...

### DIFF
--- a/src/PointCloudOctree.js
+++ b/src/PointCloudOctree.js
@@ -224,8 +224,35 @@ export class PointCloudOctree extends PointCloudTree {
 
 		let disposeListener = function () {
 			let childIndex = parseInt(geometryNode.name[geometryNode.name.length - 1]);
-			parent.sceneNode.remove(node.sceneNode);
-			parent.children[childIndex] = geometryNode;
+			//check if sceneNode Deleted maybe can be removed for permformance gain it shouldn't be null
+			if(sceneNode!==null){
+				//check if has geometry it could be removed also...
+				if(sceneNode.geometry){
+					//delete attributes solve a big memory leak...
+					for(let key in sceneNode.geometry.attributes){
+							delete sceneNode.geometry.attributes[key];
+					}
+					//dispose geometry
+					sceneNode.geometry.dispose();
+					sceneNode.geometry=undefined;
+				}
+				//check if has material, can be removed...
+				if(sceneNode.material){
+					//check if has material map, can be removed...
+					if(sceneNode.material.map){
+						sceneNode.material.map.dispose();
+						sceneNode.material.map=undefined;
+						}
+					//dispose material
+					sceneNode.material.dispose();
+					sceneNode.material=undefined;
+				}
+				//remove node from scene
+				parent.sceneNode.remove(sceneNode);
+				//set node explicitly undefined.
+				sceneNode=undefined;
+				parent.children[childIndex] = geometryNode;
+			}
 		};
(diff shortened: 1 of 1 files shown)


m-schuetz:
Hi, thanks for the PR. I've been looking a bit into this but I'm not sure it solves the problem. For one, geometry.dispose is already called in PointCloudOctreeGeometry.dispose(). This function then calls the disposeListener in PointCloudOctree, which calls dispose again. The material should not be disposed, because all sceneNodes use the same material instance. The deletes on the sceneNode instance should not be necessary since idealy, after detaching the sceneNode from the scene, the garbage collector should get rid of the sceneNode and all its properties.

HeadLessHUN:
There are memory leaks in 1.6, because it's always growing, until the point of a crash (ohh snap...). After that a browser refresh and loop begin :) 

I don't have in depth informations about the develop version, but i hope it will be get better, for now it is fairly usable. 
Sorry the duplication i just tried to follow the path of the disposal, and it was a pain in the ass for me at least who didn't work on this at all. 

I wrote a point cloud/360 image loader class, which loads different clouds/360 images, from an information json, which is generated separatelly, from the output of cyclone register 360 e57 export.

so for now we use an overview cloud which is subsampled heavily, and then we have several small clouds each around 10 million point, and gets loaded when the scan location gets opened. so it remove the need of big clouds load in the same time, and it is far better point density... (we use it for building surveys).

midnight-dev:
> There are memory leaks in 1.6, because it's always growing, until the point of a crash (ohh snap...). After that a browser refresh and loop begin :) 

I'm curious. Did this change result in a measurable difference in memory footprint over time (compared to either last release 1.6 or the latest develop branch)? If so, then the garbage controller doesn't think it should clean up some resources, which usually means there's a reference to it from somewhere. Like Markus said though, some resources are held onto for a good reason. 

I'll try to test this for myself in the near future. I'd support this work if the patch reduces memory while not having adverse side effects (like missing cached shaders, visual artifacts, high memory thrashing, etc.). There's definitely other gains to be had, but any fix is a good fix.

By the by, any memory tests for Potree should probably exclude Cesium background rendering. The library is cool, but it currently has _heaps_ of memory issues.

HeadLessHUN:
i think if you see the same render area and then change the point budget slider back and forth there shouldn't be memory increase at all. i've got succes on that side. for me it's better that way, because the clouds are that big its crashing like every 5-10 minutes. But i'm not a professional, just do it for fun. maybe it is not using the cache no matter if it's there.. iam using 1.6 because 1.7 is a little buggy for now, but recently i didn't test it.

since  then i use an another loading mechanism which is not perfect but way better i have scan location spheres, and if you double click that then its load a smaller but denser cloud which loads so much faster and allows you to see it like a 3d panorama photo, with measurements. and if you exit from that then unload completelly that cloud. i've reachd almost 30 minutes before crash on a 8 gb memory machine. the GC deletes a lot on that small memory machine.

PR #636, opened 2019-04-09
Add EPT Zstandard loader.

connormanning:
Add EPT loader for the [Zstandard](https://facebook.github.io/zstd/) compression format.  Because parsing this format for EPT consists simply of a decompression step followed by the Binary loader functionality, I've pulled the binary parsing functionality apart into `libs/ept` so it's accessible from the web workers - not sure if that's the right spot for it.  This will also make it simple to add other generic compression loaders for EPT if desired in the future.

Included to decompress this format are the emscripten compiled JS bindings from [yoshihitoh/zstd-codec](https://github.com/yoshihitoh/zstd-codec), which is MIT-licensed.

Sample dataset included at `pointclouds/lion_takanawa_ept_zst`, with viewer example at `examples/ept_zstandard.html`.


m-schuetz:
Looks good, thanks for the PR!

PR #679, opened 2019-08-01
Fix device orientation

chrislcs:
I saw you mentioned deviceorientation.html was not working in the list of high priority issues. I went to take a look and it seems like the control objects were no longer exposed in the Potree object, which the example uses to set the controls: `viewer.setNavigationMode(Potree.DeviceOrientationControls);`. Here is a quick fix for that. 

It also seems like EDL is no longer working on mobile (it does not show any point cloud with it enabled), at least not on my phone, neither using Chrome nor Samsung Internet. This does not seem to be connected to the device orientation controls though, so in this PR I just disabled EDL for this example.

### DIFF
--- a/examples/deviceorientation.html
+++ b/examples/deviceorientation.html
@@ -43,7 +43,7 @@
 
 		window.viewer = new Potree.Viewer(document.getElementById("potree_render_area"));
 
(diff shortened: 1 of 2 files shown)


m-schuetz:
Great, thanks for the fix!

PR #686, opened 2019-08-07
Matcap material support

jo-chemla:
Add backface culling and correct matcap computation for orthographic cameras.


m-schuetz:
Hi, thanks for the PR!

* Is the idea of this backface culling approach to make points black rather than discarding them entirely?
* Some lines in pointcloud.vs use spaces instead of tabs. Please convert them to tabs.

jo-chemla:
Hi again. Yes indeed, here is a screenshot - if the pointcloud has normals, a checkbox appears to allow user to enable backface culling or not. It is especially useful when doing clipping with closed objects scanned entirely (buildings, ...) and to hide what should not be visible because facing the opposite way from the camera - works in both perspective and orthographic cameras
. 
image

m-schuetz:
Great, thanks, merged! 

For the next PR, can you merge develop into your branch first? The gulp build was updated and now requires gulp 4, so you may have to update to the new one.

PR #688, opened 2019-08-09
Write a Chinese Readme

wemamawe:
i translate the readme to chinese. so I want more china person can know Potree and use Potree


chuxiaonan:
兄弟，非常好的工作👍 支持你

PR #696, opened 2019-08-24
Fix NavCube behavior when invisible

NiViggiano:
The Navigation Cube tool can change the view even when it's not visible, which seems like unintended behavior. Added conditional to fix it.


m-schuetz:
Good catch, didn't realize that!

PR #729, opened 2019-11-15
add correct access to named projections like they are used in proj4js

vvmruder:
I had problem with loading projected point clouds from EPT. It always failed with `SyntaxError: JSON.parse: unexpected character at line 1 column 1 of the JSON` for proj4js version 2.3.12 and with `Error: havn't handled ":" in keyword yet, index 5` for proj4js versions later. This leads to a not loaded point cloud and a browser dev console throwing the named errors ten of thousands of times.

So I digged deeper and it seemed that this changes repair problem with projected point clouds. So I wanted to kindly offer this for merge.

Other way of improving this, would be to store actually the proj4js object as element on the viewers `projection` attribute to not need to obtain it every time. I saw this is called heavily.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -904,7 +904,7 @@ export class Utils {
 			// if there is a projection, transform coordinates to WGS84
 			// and compute angle to north there
 
-			proj4.defs("pointcloud", projection);
+			proj4.defs("pointcloud", proj4.defs(projection));
 			const transform = proj4("pointcloud", "WGS84");
 
 			const llP1 = transform.forward(p1.toArray());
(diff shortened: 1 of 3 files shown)


m-schuetz:
In what way are projections defined in EPT? The reason it's ```proj4.defs("pointcloud", projection);``` right now is because the projection string is not a named projection, but rather a proj4 string, e.g.:

```
// EPSG:2056
let projection = "+proj=somerc +lat_0=46.95240555555556 +lon_0=7.439583333333333 +k_0=1 +x_0=2600000 +y_0=1200000 +ellps=bessel +towgs84=674.374,15.056,405.346,0,0,0,0 +units=m +no_defs";
proj4.defs("pointcloud", projection);
```

As far as I understand, the proposed change would require to specify a named projection rather than a proj4 string, like:

```
let projection = "EPSG:2056";
proj4.defs("pointcloud", projection);
```

Unfortunately that would break existing examples and potree pages. The reason that we've chosen to use proj4 strings rather than named projections is because some named projections didn't exist. A better solution that may work with both options could be something to consider for Potree 2.0.

vvmruder:
@m-schuetz Thanks for your thoughts and sorry for late response. I was consumed by other stuff these days. In my datasets the projection is defined by names AND proj string. But I might use it wrong. I understand your concerns about existing code. So we can close this PR :)

PR #751, opened 2020-01-03
Remove requirement for global gulp and rollup

lightheaded:
I did some minor improvements so that people familiar with NodeJS ecosystem can feel right at home and also newbies can follow standard practices without the need for global installations.

From commit message:

> Requiring global installation of any tools is not the best practice, as it can
> introduce situations when one project requires one version of e.g. gulp and
> another application requires another version. To combat that, npm provides a great
> way to encapsulate dependencies inside `node_modules` directory and includes all
> binaries in PATH whenever `npm ...` commands are ran.
> 
> In this commit, `rollup` was added as a dependency to `package.json` so that global
> installation is not required. Additionally, npm scripts for common operations were
> added: `npm start` for development and `npm run build` for one-time building.
> 
> README was updated to match the updates.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@ Reference: [Potree: Rendering Large Point Clouds in Web Browsers](https://www.cg
 Make sure you have [node.js](http://nodejs.org/) installed
 
 Install all dependencies, as specified in package.json: 
+
 ```bash
 npm install
 ```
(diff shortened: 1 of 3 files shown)


m-schuetz:
That makes sense, thanks!

PR #773, opened 2020-02-24
use fallback wkt in case epsg is set before using proj4.def. fixes #734

### LINKED ISSUE #734: EPT format loading fails: Unexpected token E in JSON at position 0
Hey! I am trying to load an EPT point cloud containing "srs" attributes and it fails in the latest version of Potree. This issue can be solved if the attribute is set to an empty object, but unfortunately this won't be always possible when loading third-party point clouds. I looked at the error stack trace, and this is what it produces.

```
Uncaught SyntaxError: Unexpected token E in JSON at position 0
    at JSON.parse (<anonymous>)
    at b.exports (proj4.js:formatted:3213)
    at Function.d [as defs] (proj4.js:formatted:954)
    at Function.computeAzimuth (potree.js:3064)
    at Viewer.<anonymous> (potree.js:16967)
    at Viewer.dispatchEvent (potree.js:98)
    at Viewer.update (potree.js:28247)
    at index.js:214
```

The issue is with the `proj4.js` code that looks like this.

```
b.exports = function(a, b) {
                var c = JSON.parse(("," + a).replace(/\s*\,\s*([A-Z_0-9]+?)(\[)/g, ',["$1",').slice(1).replace(/\s*\,\s*([A-Z_0-9]+?)\]/g, ',"$1"]').replace(/,\["VERTCS".+/, ""))
                  , d = c.shift()
                  , f = c.shift();
```

Here, the `a` variable is `EPSG:4326` which is not a valid JSON, so parsing it will throw an error. Has anyone else run into this issue before? Are you able to help @connormanning? Thanks!

luipir:
This PR should file (at least) EPT sources when ept.json files contain srs like that set in: https://github.com/potree/potree/issues/734
e.g. when both epsg code and wkt are set.

this PR was born from https://github.com/connormanning/entwine/issues/224 and does NOT solve the case when wkt is a compound srs not managed by proj4js


luipir:
rebased

luipir:
sorry @midnight-dev looking for time to come back on this PR

PR #791, opened 2020-04-15
Fixing azimuth calculations on EPT sources and adding EPSG defs

mccarthyryanc:
I have some EPT sources where the `projection` string that gets defined is an EPSG code string, like `EPSG:6339`. So I added a check to use that, instead of a prjo4 string in transform step when calculating azimuth.

I also added some proj4 definitions that I use frequently. Hopefully `map.js` is an okay place to do this. But happy to move or remove if necessary.


m-schuetz:
Seems fine, thanks for the PR!

PR #795, opened 2020-04-21
NavigationCube position correction

ilisalis:
Correction of NavigationCube display for all display modes (missing for standard quality, EDL disabled). 
Placement at the bottom to avoid overlapping with the compass.


m-schuetz:
Thanks for the PR. It's quite a lot to take in at once and I'll add some comments/questions in the PR.

PR #800, opened 2020-04-28
Resolve Potree Crash When Selecting Level of Detail - Fixes #798

### LINKED ISSUE #798: Crash in Examples: Pointclouds with Fixed Point Sizing and LOD attribute
Currently, all Potree examples with point-clouds crash the browser (Chrome or Firefox latest) when:

- A point-cloud is selected. e.g sigeom.sa in the [Annotations example](http://potree.org/potree/examples/annotations.html)
- Under properties, the point sizing is set to Fixed
- Under Attribute, Level of Detail is selected   

The problem seems to lie in the shader code (see screenshot)
image
image

midnight-dev:
When using fixed point sizes and filtering for Level of Detail (LOD), Potree would suffer a fatal crash. 
This went unnoticed for a while because adaptive size is pretty common. Closes #798 👍 

The cause was due to inconsistent references in the point cloud vertex shader.
`color_type_lod` vs `color_type_level_of_detail`

I would have preferred using the shorter of the two, but that would have created other issues.

@m-schuetz Now that this PR is cleaned up, please review it when you get a chance.

### DIFF
--- a/src/materials/shaders/pointcloud.vs
+++ b/src/materials/shaders/pointcloud.vs
@@ -149,7 +149,7 @@ float round(float number){
 // OCTREE
 // ---------------------
 
-#if (defined(adaptive_point_size) || defined(color_type_lod)) && defined(tree_type_octree)
+#if (defined(adaptive_point_size) || defined(color_type_level_of_detail)) && defined(tree_type_octree)
 /**
  * number of 1-bits up to inclusive index position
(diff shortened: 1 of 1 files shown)


m-schuetz:
Hi, thanks for the PR, I'll check it out tomorrow!

m-schuetz:
Works! Thanks!

PR #808, opened 2020-05-11
WebGL Improvements to Fix #807

### LINKED ISSUE #807: Warnings on Initial Load - WebGL & ThreeJS
This is a long-standing issue. Documenting it here. Upon loading Potree, there are several warnings that appear in the console of Firefox. There are also warnings in Chrome but to a lesser extent.

image

This is due to WebGL extensions being forced on implicitly rather than being used after enabling them. Namely `WEBGL_color_buffer_float` - there are some other extensions that are sometimes used, but generally not on first load.

There is also a warning from ThreeJS regarding a Matrix3 getInverse() problem. It's unable to inverse the matrix because it's an invalid matrix filled with all 0 values.

midnight-dev:
This series of commits should close #807 in time for Potree 1.7 👌 

Enabling the float color buffer on the viewer and profile was simple. Also cleaned up a duplicate extension call for frag depth in profile.js.

The tricky part was pinning down the cause for this:
> THREE.Matrix3: .getInverse() can't invert matrix, determinant is 0

This comes from when the profile is first rendered, and it is always rendered upon initialization whether it has a reason to display or not due to the `resize` event listener. And as you can guess, this means that _every_ time the window is resized in some way (even an orientation change), the profile would re-render. Playing with the window size for approximately 1 second on my laptop resulted in over 100 warning messages.

image

So now, before we allow for a `resize` event to render the profile window, we check to see if the profile is actually open (enabled). I haven't tested this against all possible use cases with the profile feature, but it works flawlessly on everything that I use it for. Since having the profile window open should _always_ make the `enabled` flag true, reason follows that this ought to be free from unwanted side effects. While the profile window is closed, we no longer waste CPU cycles on the render logic during init and any resize.

Using these commits on a localhost server, most of the warnings are no longer present on load. The one about lazy initialization can't be helped right now.

image

While I was fixing up profile.js, I noticed `setProfile()` doesn't actually do anything inside the `ProfileWindow` class. It's never called anywhere. It's the `ProfileWindowController` class that has a working `setProfile()` method. Since that's the one that is used, we can remove the accidental orphan method from `ProfileWindow`.

### DIFF
--- a/src/viewer/profile.js
+++ b/src/viewer/profile.js
@@ -697,10 +697,6 @@ export class ProfileWindow extends EventDispatcher {
 			.call(this.yAxis);
 	}
 
-	setProfile (profile) {
-		this.render();
-	}
-
 	addPoints (pointcloud, points) {
 
 		let entry = this.pointclouds.get(pointcloud);
--- a/src/viewer/profile.js
+++ b/src/viewer/profile.js
@@ -614,11 +614,7 @@ export class ProfileWindow extends EventDispatcher {
 
 
 		{
-			// See viewer.js for note on extensions
 			let gl = this.renderer.getContext();
-			gl.getExtension('EXT_frag_depth');
-			gl.getExtension('WEBGL_depth_texture');
-			gl.getExtension('WEBGL_color_buffer_float');
 
 			let extVAO = gl.getExtension('OES_vertex_array_object');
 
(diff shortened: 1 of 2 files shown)


m-schuetz:
Thanks for the PR. I've also removed the other extension initializers in profile.js, since profile rendering doesn't require them anyway.

PR #809, opened 2020-05-12
Security Fixes, Prepare Package.json for Version 1.7

midnight-dev:
There are a plethora of security fixes that have been released for Potree's NPM dependencies. 

image

This brings down the vulnerabilities to a single (low severity) prototype pollution issue of `yargs-parser`. This seems to be something that Gulp makes brief use of. I would update this too, manually, however the current dependency requirement is for version 5.x.x, and there is only 5.0.0 available. The next newest version is 6.0.0, which is a major version change that might have breaking changes. Funny enough, that version is also vulnerable though. This script would have to be updated to version 13+ in order to not be affected. Compatibility is definitely not guaranteed for such a heroic upgrade. It would be best to let the Gulp crew manage this and update when they have verified compatibility with a later version of the yargs script.

On a related note, `gulp-util` has been deprecated and unmaintained for years. Its authors strongly urged users to migrate to other options 3 years ago. Since it's not used by anything in the Potree suite (nor used by any dependencies), it can and should be removed. This helps reduce security and maintainability problems going forward. If someone needs it for an edge case, it can still be installed on-demand with `npm install gulp-util`. For more info, check the official distribution site: https://www.npmjs.com/package/gulp-util

Lastly, this PR updates the `package.json` file with a few keywords and syncs the version to the upcoming Potree 1.7 (instead of version 0.0.1) 🎉 

TODO: 

- [x] Update Potree JavaScript version. There is the constant version object, but also references elsewhere like in `PointCloudEptGeometry`. 
- [x] Might want to check PotreeConverter soon in case there are 1.6-specific references or conditionals.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,18 +1,18 @@
 {
   "name": "potree",
-  "version": "1.7.0",
+  "version": "0.0.1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
     "@types/estree": {
-      "version": "0.0.44",
-      "resolved": "https://registry.npmjs.org/@types/estree/-/estree-0.0.44.tgz",
-      "integrity": "sha512-iaIVzr+w2ZJ5HkidlZ3EJM8VTZb2MJLCjw3V+505yVts0gRC4UMvjw0d1HPtGqI/HQC/KdsYtayfzl+AXY2R8g=="
+      "version": "0.0.42",
+      "resolved": "https://registry.npmjs.org/@types/estree/-/estree-0.0.42.tgz",
+      "integrity": "sha512-K1DPVvnBCPxzD+G51/cxVIoc2X8uUVl1zpJeE6iKcgHMj4+tbat5Xu4TjV7v2QSDbIeAfLi2hIk+u2+s0MlpUQ=="
     },
     "@types/node": {
-      "version": "13.13.5",
-      "resolved": "https://registry.npmjs.org/@types/node/-/node-13.13.5.tgz",
-      "integrity": "sha512-3ySmiBYJPqgjiHA7oEaIo2Rzz0HrOZ7yrNO5HWyaE5q0lQ3BppDZ3N53Miz8bw2I7gh1/zir2MGVZBvpb1zq9g=="
+      "version": "13.7.4",
+      "resolved": "https://registry.npmjs.org/@types/node/-/node-13.7.4.tgz",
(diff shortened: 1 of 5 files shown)


midnight-dev:
@m-schuetz Ready to merge if you approve. This brings the NPM package info up to 1.7, fixes the security issues in the dependencies, and finally bumps the Potree JS to 1.7. Made use of the suffix so that it currently shows as `Potree 1.7beta` in the console. If you don't want the beta suffix, just blap it from this PR. Made it under a separate commit (06f965f) so it could be removed or reverted easily.

Just to be thorough, I checked PotreeConverter and didn't find anything version-number specific to Potree other than assuring that the format is above version 1.4. There was also [this commit from 2015](https://github.com/potree/PotreeConverter/commit/c16f3892470c2fba543ddffd345fd4bb3c0f271b) adding in a reference to 1.6 in `PotreeWriter`'s CloudJS stuff, but it's changed in the years since then. However, I'm not setup for C++ on my laptop at the moment, so it's not like I stepped through with a debugger. I performed old fashioned text searches to see if anything obvious stuck out. (I'll get more acquainted with Converter when I'm back at my usual workstation with Visual Studio on it.)

Awaiting your review. Cheers o/

m-schuetz:
Hi, thanks for the PR. Could you revert the changes to the package-lock file? Since it's a rather security sensitive part, I'd like to update that one myself. Perhaps you could tell me which steps/commands are necessary to do so?

midnight-dev:
Sure. Reverted `package-lock.json`. This also knocks out the version bump.

### To see the security issues

Checkout / clone dev branch. After running `npm install`, you may be notified of the many security issues automatically. If not, run `npm audit`.

### To fix these, let's regenerate the lock file

Delete your lock file (or rename it, whichever is more convenient). Delete your `node_modules` directory, as we'll want to get fresh versions. Important part - make sure your `package.json` file has the `version` key set if you want your lock file to automatically include it instead of it using 0.0.1. Also important, remove `gulp-util` as a dependency if you want to get a lock file that has parity with the commit I made. See the reasons in my PR why this dependency isn't needed and shouldn't be included.

With the package lock file and node modules removed, run a normal `npm install`. NPM will fetch the most up-to-date dependencies that your package JSON asks for. After a few moments, you should have a shiny new lock file that will resolve all of the severe security issues. **But just to be sure,** run another NPM audit to see what, if anything, needs attention. When I had built the lock file in this PR, only a single low severity security issue remained from `yargs-parser` because of Gulp.

Lemme know if there are any other concerns.

m-schuetz:
nice, package-lock looking good now. thanks.

PR #816, opened 2020-05-16
Fixes for Matrix3 / Matrix4 .getInverse() Warnings - Clipping & Height Profiling

### LINKED ISSUE #814: Console Warning: THREE.Matrix3: .getInverse() can't invert matrix
Potree will sometimes litter the console with these warnings.

> THREE.Matrix3: .getInverse() can't invert matrix, determinant is 0

image

This typically occurs during user action, such as measuring or changing view, but it varies with point clouds, camera position, and marker position since all of these variables are factors in matrix calculations for internal Potree methods. This can be reproduced using the following steps:

1. Open the [basic viewer](http://potree.org/potree/examples/viewer.html) example.
2. Take a height profile with several points
3. Open the profile window (this can be closed or left open)
4. Move the profile points around

I haven't figured out the exact method to the madness, but sometimes there are very few Matrix3 issues, sometimes there are many on every adjustment or orientation change. Camera position and points matter, but maybe FOV as well. The problem occurs in ThreeJS itself when it's trying to inverse a bad matrix.

I intend on providing more of a detailed write-up for the cousin of this warning, the Matrix4 `.getInverse` warning. That one wastes significantly more CPU than this one. But both are problematic.

This will be fixed soon, but I wanted to document it here for anyone searching for the problem. Might have to reference this in the future tackling related matters, too.

### LINKED ISSUE #815: Console Warning: THREE.Matrix4: .getInverse() can't invert matrix
Under certain conditions, Potree will emit warnings regarding Matrix4 inversion. These come specifically from ThreeJS and look like the following:

> THREE.Matrix4: .getInverse() can't invert matrix, determinant is 0

image

This typically occurs upon marking areas that need to be clipped or profiled. It sometimes happens and other times doesn't. I believe it's based on several factors including point cloud & node matrices and marker position. Point cloud density or loaded points may also play a part by increasing or decreasing the usable precision of surface area when placing markers. I imagine this means that some users see this happening constantly while others hardly see it at all due to everyone working with different point clouds of different makeups.

The primary cause of this is as the warning message says. There's an error while attempting to invert a matrix because the determinant is 0. What does that mean? The math behind inverting a 4x4 matrix is a pain in the ass, and I'm not a professor of linear algebra, so I'll skip the juicy bits. Some basic arithmetic is done on the elements of the array, namely addition and multiplication with some subtraction too. Let's take a look to see how ThreeJS determines the determinant: 

```js
var te = this.elements,
	me = m.elements,  // matrix elements passed into the .getInverse() method

	n11 = me[ 0 ], n21 = me[ 1 ], n31 = me[ 2 ], n41 = me[ 3 ],
	n12 = me[ 4 ], n22 = me[ 5 ], n32 = me[ 6 ], n42 = me[ 7 ],
	n13 = me[ 8 ], n23 = me[ 9 ], n33 = me[ 10 ], n43 = me[ 11 ],
	n14 = me[ 12 ], n24 = me[ 13 ], n34 = me[ 14 ], n44 = me[ 15 ],

	t11 = n23 * n34 * n42 - n24 * n33 * n42 + n24 * n32 * n43 - n22 * n34 * n43 - n23 * n32 * n44 + n22 * n33 * n44,
	t12 = n14 * n33 * n42 - n13 * n34 * n42 - n14 * n32 * n43 + n12 * n34 * n43 + n13 * n32 * n44 - n12 * n33 * n44,
	t13 = n13 * n24 * n42 - n14 * n23 * n42 + n14 * n22 * n43 - n12 * n24 * n43 - n13 * n22 * n44 + n12 * n23 * n44,
	t14 = n14 * n23 * n32 - n13 * n24 * n32 - n14 * n22 * n33 + n12 * n24 * n33 + n13 * n22 * n34 - n12 * n23 * n34;

var det = n11 * t11 + n21 * t12 + n31 * t13 + n41 * t14;  
// This is just the determinant, plenty of math after this point too
```

That's a lot of arithmetic, but remember there are 16 values in a Matrix4. That leads to a lot of steps to eventually get the the determinant which you can then use in further math. The key here is that multiply operation. More on that in a second. What if there's an invalid Matrix4 that doesn't play nice with the Markovs? 🌵 

```js
[
    0, 0, 0, 0,
    0, 0, 0, 0,
    0, 0, 0, 0,
    0, 0, 0, 0
]
```

Well, if they're all zeros, you're going to have a bad time. 
It won't invert, and you'll only be given an identity matrix as a consolation prize.

```js
[
    1, 0, 0, 0,
    0, 1, 0, 0,
    0, 0, 1, 0,
    0, 0, 0, 1
]
```

This will chooch the affine jibberjabber, satisfy the eigenvectsors, and pull a sick 360 kickflip while winking at you, but that doesn't help in any meaningful way. It's just trying to give back _something_ when it can't actually invert. Most of the time, the matrices are not full of all zeros. However, when a certain 4 are 0, the determinant will always be 0 too. Take the following - a real value I just pulled from the debugger when an inversion failed.

> 0: 0
> 1: 0
> 2: 0
> 3: 0
> 4: -999999.9949999999
> 5: 3.3306690738754696e-10
> 6: 99.99999949999999
> 7: 0
> 8: 0.00025799999871
> 9: -1.748276003132876e-20
> 10: 2.5799999871000003
> 11: 0
> 12: 589675.9697493133
> 13: 231467.90974984743
> 14: 758.9950016784668
> 15: 1

Although the matrix has plenty of suitable values, the top row elements will be multiplied against other values before an inverse can be made. The result is that everything goes 0, meaning there can't be an inversion. ThreeJS only has one course of action it can take, and that's to return an identity Matrix4. During the determinant calculation, our last line might as well read:

```js
// Spoiler alert: it's going to be 0
det = (0 * someNumber) + (0 * someNumber) + (0 * someNumber) + (0 * someNumber); 
```

If you're picking up what I'm laying down, then you may not be shocked by my soon-to-come pull request that addresses this issue. The fix for this is going to reduce the unwanted warnings in our consoles and also keep things more efficient. The idea being that there will no longer be CPU cycles wasted preparing a determinant that will ultimately be 0.

A couple things I should note before wrapping up. This is strongly related to the previous issue #814 but that one is less wasteful than this one. There aren't as many steps in reverting a 3x3 matrix as there are for a 4x4 matrix. Shocking, I know. At a ratio of 16:9, the Matrix4 has nearly twice as many elements as the Matrix3 and calculating the inverse is more complicated. So this is the ugly cousin of the two issues.

The commits I wrote to fix the matrix errors aren't a complete fix, as I still don't know why it comes up with zeros in the top row so often. There's an origin that I'm not seeing. If anyone has a solid lead on what's going on with the matrix 0s, comment here or link this issue so we can have a complete record for future engine rewrites and other open source efforts 👌

midnight-dev:
This used to be the result of getting a simple height profile or clipping something - anywhere from dozens to hundreds of matrix inversion fails resulting in more processor usage and a polluted console log.

image

Excerpts from the commit logs.

Closes #814 **

### THREE.Matrix3: .getInverse() can't invert matrix, determinant is 0

> This patches the condition where d3 scaleX invert results in 0 radius.
> 
> When the point's scale is set to 0, 0, 0, the position becomes 0, 0, 0,
> and ultimately, all of the matrix elements become 0, which is invalid.

Closes #815 **

### THREE.Matrix4: .getInverse() can't invert matrix, determinant is 0

> This patches the condition where first 4 Matrix4 elements are 0.
> 
> This sometimes occurs with measurements. These elements will always be factored with the other elements when getInverse() is called, meaning the result will be 0 (an invalid state with no inverse).
> In these cases, ThreeJS will return an identity Matrix4. We can sidestep a bunch of math, saving CPU cycles, by checking these 4 values. If all are 0 or don't exist, the determinant will always be 0.

Basically, the big save here is that we are providing the same behavior (returning an identity Matrix4 when the determinant is going to be 0) but without wasting computation on finding that determinant value.

** These fixes do not address the origin issue where there are sometimes bad matrices for markers and an invalid state for d3 scaleX. We should want to investigate how to resolve the root causes in the near future.

I'm also going to be probing to see if I can get matrix warnings under other conditions, but these fixes are substantial enough on their own to include them on their own.

### DIFF
--- a/src/viewer/profile.js
+++ b/src/viewer/profile.js
@@ -878,6 +878,11 @@ export class ProfileWindow extends EventDispatcher {
 		pRenderer.render(profileScene, camera, null);
 
 		let radius = Math.abs(scaleX.invert(0) - scaleX.invert(5));
+
+		if (radius === 0) { 	// If 0, all matrix elements will be 0 and therefore invalid (cannot be inversed)
+			radius = 0.000000001;
+		}
+
 		pickSphere.scale.set(radius, radius, radius);
 		pickSphere.visible = true;
 		renderer.render(scene, camera);
--- a/src/viewer/profile.js
+++ b/src/viewer/profile.js
@@ -879,12 +879,13 @@ export class ProfileWindow extends EventDispatcher {
 
 		let radius = Math.abs(scaleX.invert(0) - scaleX.invert(5));
 
-		if (radius === 0) { 	// If 0, all matrix elements will be 0 and therefore invalid (cannot be inversed)
-			radius = 0.000000001;
+		if(radius === 0){
+			pickSphere.visible = false;
+		}else{
+			pickSphere.scale.set(radius, radius, radius);
+			pickSphere.visible = true;
 		}
-
-		pickSphere.scale.set(radius, radius, radius);
-		pickSphere.visible = true;
+		
 		renderer.render(scene, camera);
 
 		this.requestScaleUpdate();
--- a/src/viewer/profile.js
+++ b/src/viewer/profile.js
@@ -879,9 +879,9 @@ export class ProfileWindow extends EventDispatcher {
 
 		let radius = Math.abs(scaleX.invert(0) - scaleX.invert(5));
 
-		if(radius === 0){
+		if (radius === 0) {
 			pickSphere.visible = false;
-		}else{
+		} else {
 			pickSphere.scale.set(radius, radius, radius);
 			pickSphere.visible = true;
 		}
--- a/src/viewer/viewer.js
+++ b/src/viewer/viewer.js
@@ -1760,8 +1760,12 @@ export class Viewer extends EventDispatcher{
 			
 			let clipBoxes = boxes.map( box => {
 				box.updateMatrixWorld();
-				let boxInverse = new THREE.Matrix4().getInverse(box.matrixWorld);
-				let boxPosition = box.getWorldPosition(new THREE.Vector3());
+
+				// Fix for getInverse "determinant is 0" - saves resources
+				const matrixMultipliersZero = !box.matrixWorld.elements[0] && !box.matrixWorld.elements[1] && !box.matrixWorld.elements[2] && !box.matrixWorld.elements[3];
+				const boxInverse = matrixMultipliersZero ? new THREE.Matrix4() : new THREE.Matrix4().getInverse(box.matrixWorld);
+
+				const boxPosition = box.getWorldPosition(new THREE.Vector3());
 				return {box: box, inverse: boxInverse, position: boxPosition};
 			});
 
--- a/src/viewer/viewer.js
+++ b/src/viewer/viewer.js
@@ -1758,7 +1758,8 @@ export class Viewer extends EventDispatcher{
 				boxes.push(...profile.boxes);
 			}
 			
-			let degenerate = (box) => box.matrixWorld.determinant() !== 0;
+			// Needed for .getInverse(), pre-empt a determinant of 0, see #814
+			let degenerate = (box) => box.matrixWorld.elements[0] || box.matrixWorld.elements[1] || box.matrixWorld.elements[2] || box.matrixWorld.elements[3];
 			
 			let clipBoxes = boxes.filter(degenerate).map( box => {
 				box.updateMatrixWorld();
--- a/src/viewer/viewer.js
+++ b/src/viewer/viewer.js
@@ -1758,14 +1758,14 @@ export class Viewer extends EventDispatcher{
 				boxes.push(...profile.boxes);
 			}
 			
-			let clipBoxes = boxes.map( box => {
+			let degenerate = (box) => box.matrixWorld.determinant() !== 0;
+			
+			let clipBoxes = boxes.filter(degenerate).map( box => {
 				box.updateMatrixWorld();
+				
+				let boxInverse = new THREE.Matrix4().getInverse(box.matrixWorld);
+				let boxPosition = box.getWorldPosition(new THREE.Vector3());
 
-				// Fix for getInverse "determinant is 0" - saves resources
-				const matrixMultipliersZero = !box.matrixWorld.elements[0] && !box.matrixWorld.elements[1] && !box.matrixWorld.elements[2] && !box.matrixWorld.elements[3];
-				const boxInverse = matrixMultipliersZero ? new THREE.Matrix4() : new THREE.Matrix4().getInverse(box.matrixWorld);
-
-				const boxPosition = box.getWorldPosition(new THREE.Vector3());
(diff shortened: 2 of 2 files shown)


m-schuetz:
Thanks for identifying the issue, I'll modify the changes a bit.

m-schuetz:
Let me know if you find issues with the changes.

midnight-dev:
Looks good. I'll test it to be sure.

midnight-dev:
@m-schuetz I like it it, and the warnings are still fixed. But we're now computing the determinant twice for Matrix4s. 

Although it's less attractive, it would be more performant to check if the top 4 elements are anything but 0. How about splitting the difference so the 4 top elements are checked during the degenerate filter? Otherwise this will **_always_** occur on every update in addition to the inversion (code from Three's .determinant() function):

```js
var te = this.elements;

var n11 = te[ 0 ], n12 = te[ 4 ], n13 = te[ 8 ], n14 = te[ 12 ];
var n21 = te[ 1 ], n22 = te[ 5 ], n23 = te[ 9 ], n24 = te[ 13 ];
var n31 = te[ 2 ], n32 = te[ 6 ], n33 = te[ 10 ], n34 = te[ 14 ];
var n41 = te[ 3 ], n42 = te[ 7 ], n43 = te[ 11 ], n44 = te[ 15 ];

//TODO: make this more efficient
//( based on http://www.euclideanspace.com/maths/algebra/matrix/functions/inverse/fourD/index.htm )

return (
	n41 * (
		+ n14 * n23 * n32
			- n13 * n24 * n32
			- n14 * n22 * n33
			+ n12 * n24 * n33
			+ n13 * n22 * n34
			- n12 * n23 * n34
	) +
	n42 * (
		+ n11 * n23 * n34
			- n11 * n24 * n33
			+ n14 * n21 * n33
			- n13 * n21 * n34
			+ n13 * n24 * n31
			- n14 * n23 * n31
	) +
	n43 * (
		+ n11 * n24 * n32
			- n11 * n22 * n34
			- n14 * n21 * n32
			+ n12 * n21 * n34
			+ n14 * n22 * n31
			- n12 * n24 * n31
	) +
	n44 * (
		- n13 * n22 * n31
			- n11 * n23 * n32
			+ n11 * n22 * n33
			+ n13 * n21 * n32
			- n12 * n21 * n33
			+ n12 * n23 * n31
	)

);
```

That's a lot of unnecessary allocations and arithmetic. The minor alteration would mean we only compute the determinant once during the inversion. Check the minor tweak to see what you think.

m-schuetz:
I don't think it's much of a performance bottleneck since the number of clip boxes & volumes is very limited anyway. I'd rather call .determinant() for now, than verify that checking if elements 0, 1, 2 and 3 are zero has the same effect.

midnight-dev:
Alrighty. Your call. I'll revert.

m-schuetz:
thanks!

PR #820, opened 2020-05-18
Gulp hosting + ios/xaml support

rooi:
I found that the gulp server was not accessible from other machines any longer. Adding host 0.0.0.0 option enables it.

Adding a webgl2 check allows greater browser support (tested on ios and xaml webview)


rooi:
The last three commits were not intended in this pull request. Sorry, closing it

PR #821, opened 2020-05-18
Properties fix

pierotofy:
Currently the properties panel throws an exception when users try to change the point cloud attributes (rgba, color, elevation, etc.) due to the attribute var not always being set to a proper object (sometimes it's null). This trips the code when `attribute.name` is accessed.


pierotofy:
Thanks for the review! I haven't tested with the examples, but I experienced this while integrating Potree in https://github.com/OpenDroneMap/WebODM

midnight-dev:
Commit 24e9d0877d6089bc05dbaf3efd64ac6057fbfe06 fixes this in a similar fashion. This can be closed as it now conflicts with the new loader format anyway.

m-schuetz:
Hi, thanks for the suggestion, this has already bin fixed in a recent commit, quite similarly actually.

PR #828, opened 2020-05-27
Quality of Life Improvements

midnight-dev:
This pull request covers several areas, but they're all focused on making dev work on Potree easier, faster, and more robust.

I created a `developer.html` example, making use of unminified sources (JS & CSS). This has some additional comments which may help newcomers. Added a framerate throttle, though Headless beat me to it. Decide which you like better. The limit is 60 FPS by default or whatever is passed into the Viewer() constructor, which allows this example to consume **only 10-15% of my CPU & GPU on my laptop**. And with only 100k points loaded initially, it's very quick to load / reload during testing. For managing the FPS throttle, the following code will be relevant:

```js
changeFPS( limit );  // The limit is in Hz, should be in the range of 0-300
pauseToggle();  // Pauses and resumes render operations for the WebGL context
// NOTE: These belong to the Viewer object but they can be used globally

// You can also pass in the setting through the arguments object like so
let viewer = new Potree.Viewer(document.getElementById("potree_render_area"), {
	fpsTarget: 30
});
```

@mrdoob's Stats frames counter is now updated along with minified & map files. Frame times and FPS are now correctly counted instead of staying pinned to the theoretical maximum that the browser can supply. That said, vsync is still a thing, so you may notice larger dips on laptops vs desktops with vertical sync enabled. It's placement can also be overridden with a simple CSS selector: `#fpsStats`.

image

I enabled the bounding box visuals by default so devs can get a more intuitive understanding of where most of the point density is and how POV affects what is culled or loaded. Similarly, LOD is the default material with a low opacity to demonstrate that there are plenty of options for loading clouds.

image

This also includes the developer source of jQuery UI (unmangled), which is helpful when trying to extend the existing UI, especially the profile window. And you don't _have to_ blacklist these kinds of scripts now because they're in a readable state.

image

There are more NPM tasks, specifically for starting the dev server without forcing a rebuild & alternate port tasks for those who want to run multiple instances or roll with different ports. While I was in the Gulp config, I tidied up and changed some variables to constants to practice good hygiene.

Configured Firefox & Chrome debugging launches for VSCode - really great for multiple monitors. The defaults are set assuming the `.vscode` directory is in the root of the workspace.

image

Debugging inside the editor requires either the extension for FF or Chrome. Pick your poison, just read the instructions. A bit of setup is necessary but worth it. I may make a markdown document (or a wiki entry, if allowed) about this if others are interested.

VSCode config no longer takes away the minimap or makes font really large (on 1080p displays), but formatting-relevant settings are maintained. This has made a huge difference for me; twice the screen real estate & the minimap made me more productive. Allowing blank space to be hidden also helps, because I have terrible astigmatism. I hope this helps others, too, as it will override fewer custom user preferences.

### Before
image

### After
image

For reference, I left the settings as comments for easier reversion or trial experiments. The disabled settings would be better handled at global/user settings for VSCode, as they are mostly cosmetic. The changes shouldn't compromise the maintainability of the Potree codebase. For instance, there's no reason why a minimap of code would diminish the quality of code. It's just a developer preference. Having these settings will automatically override the user's settings as it's more localized when opening a clone (including when reviewing a pull request). Some settings are 2 years old and could use a review.

image

Formatting on type, however, could lead to a formatting nightmare on every diff. For those who don't have [Shader languages support for VS Code](https://github.com/stef-levesque/vscode-shader) installed, C++ is indeed the closest fit for shader files - but you should really checkout that extension. Things like this that maintain quality of life and quality of the codebase are good and should override user settings if they plan to merge into Potree 1.x - my thoughts on the matter.

### Acknowledgements
@HeadLessHUN - Beat me to the commit for FPS limiter #635 (Long PR, still open) f0a2f53cb9597cd9a5598bf3d1cbd2f074bfc4c2
@rooi - Added network access via host 0.0.0.0 #820 (Big PR because Cesium, self-closed) c8a4ef5bed66e02c69ecfd73e43c05061a2da822
I think I saw 0.0.0.0 mentioned elsewhere but I can't find it right now. Apologies if I left you out.

I recommend viewing commits separately if formatting tweaks cloud the review. I decided to overwrite a couple earlier commits but I didn't want to force push the branch again. Might clean that up later. I'm going to submit this as a draft first to hopefully get some feedback before a formal review.

Cheers o/


midnight-dev:
@m-schuetz Cleaned up commits. Squashed some minor adjustments, no conflicts, and everything works as expected.

Ready for review.

PR #837, opened 2020-06-19
Annotation Panel Edition

ilisalis:
Change in the annotation panel:
- edition following the click of a button;
- description: display the html result and edit the html code;
- add the removing of annotation (in cascade);
- add the recording of camera position and orientation;
- add the possibility to deplace the annotation (via AnnotationTool.js);
- add the hierarchization of annotations.

Exemple:
Capture d’écran (11)

In the potree:develop branch, there is an issue when the annotation title is display in html mode (elTitle.html(annotation.title);) ; the title in the principal window disappears. My proposal is to use text mode.

### DIFF
--- a/resources/lang/en/translation.json
+++ b/resources/lang/en/translation.json
@@ -80,6 +80,12 @@
 		"show": "show volume"
 	},
 	"annotations": {
+		"annotation_position": "Position",
+		"annotation_title": "Title",
+		"annotation_description": "Description",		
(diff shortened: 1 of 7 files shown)


shaasta:
Very cool, thanks @ilisalis !!! I tried to implement these updates on my end, but they seem to conflict with LoadProject for me.. Are you able to export / import potree json files with this modification?

PR #854, opened 2020-07-24
Add postinstall script and main property.

iamscottcox:
- Runs the build script after `npm install`
- References a main file so that `import { x } from 'potree'` can be used.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,9 +15,11 @@
     "browser",
     "tool"
   ],
+  "main": "./build/potree/potree.js",
   "scripts": {
     "start": "gulp watch",
-    "build": "gulp build pack"
+    "build": "gulp build pack",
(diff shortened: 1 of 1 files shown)


m-schuetz:
Thanks! Running the build on install makes sense. How exactly do you leverage the "main" attribute, though? Does it mean ```import { x } from 'potree'``` can be used in any html file, e.g. even in ./examples/viewer.html, or would this only work for html files in the project root folder?

iamscottcox:
As far as I'm aware this will only work if you're importing from another ES Module. The majority of exports are declared in /build/potree/potree.js, so it seemed to be a prime candidate for the "main" property. It should allow anyone wanting to consume this as an ES Module to do so with a few shims on their side (setting THREE.js and other dep declarations in their webpack.config for example).

PR #856, opened 2020-07-29
Panorama mouseup to enter, no unintentional hover

TaylorDale:
Changed the triggering event to view a panorama to Mouseup, and added code to cancel highlighting if the user is actually dragging the view around, at which point a hover is undesirable.

This made the experience with the viewer a lot smoother, for example, we have a project with many panorama spheres. In its old form, clicking down to pan the view around would result in an unintentional "focus" on the panorama image. With this new method, the user has to very explicitly want to focus on the image. 

Also seems to prevent a few other strange bugs that would happen when accidentally focusing into an image, making the pano spheres disappear and the red "hover" sphere being made a permanent background after exiting the pano view.

### DIFF
--- a/src/modules/Images360/Images360.js
+++ b/src/modules/Images360/Images360.js
@@ -73,7 +73,7 @@ export class Images360 extends EventDispatcher{
 		});
 		viewer.inputHandler.addInputListener(this);
 
-		this.addEventListener("mousedown", () => {
+		this.addEventListener("mouseup", () => {
 			if(currentlyHovered){
 				this.focus(currentlyHovered.image360);
 			}
@@ -220,8 +220,19 @@ export class Images360 extends EventDispatcher{
 		// let tStart = performance.now();
 		raycaster.ray.copy(ray);
 		let intersections = raycaster.intersectObjects(this.node.children);
-
-		if(intersections.length === 0){
+		
+		// Make sure the drag is actually a drag, not just a click
+		let isDragging = false;
+		const { drag } = viewer.inputHandler;
+		if (drag) {
+			const dragThreshold = 10;
+			if (Math.abs(Math.hypot(drag.start.x - drag.end.x, drag.start.y - drag.end.y)) > dragThreshold) {
+				isDragging = true;
+			}
+		}
+		
+		// Don't highlight if no intersection or user is actually draggging
+		if(intersections.length === 0 || isDragging){
 			// label.visible = false;
 
 			return;


midnight-dev:
@TaylorDale Hi there. Thanks again for this addition. Could you take a look at my review question when you get a chance?

TaylorDale:
@midnight-dev No worries, I can't seem to find any comments though? Thanks

midnight-dev:
Just looked online while signed out, and it is indeed not visible. Which is a bummer. Maybe I can't review pulls due to permissions. Either way... About 29 days ago, I posed a question regarding the drag detection math. Not sure if I can highlight specific lines here like in a review question, but I think the lines should be obvious. Below is a quote: 

> I'm glad to see this PR. It's the kind of thing that could be easily missed during initial development yet I can see how this would be irritating. Still blown away by the feature of 360 images, but I never noticed the action was listening to mousedown. While I'm not able to test it, I'm curious if you can explain the motivation behind this formula. Not familiar with dragging recognition in the least. Is this a well-known screenspace arithmetic? Worried if start/end subtraction is in the right order. Also, if for some reason the user makes a perfect horizontal/vertical movement, will the hypotenuse still come back with something above 0? That is, if the (x, y) differences turn out to be (0, 25), that would be a flat line. It's been quite a few years since I had an algebra class. Would the `hypot` method want to return the C from A² + B² = C² or would it see that as a line and return 0 or `NaN`?

> Good catch to prevent accidental activation. Good quality of life bump toward a better UX. Just wish I had my equipment to test this beauty.

Thanks for the nice tweak 👍

TaylorDale:
Ah I see no I wasn't aware you'd responded at all! 
To answer your questions: 

- The formula is just to get the magnitude of the user's mouse/touch input. If the amount of pixels moved is greater than the constant (perhaps some thought needs to be put into where/how that is configured too, since it is just currently a made up number in the middle of the script) then it is not possible that the user is trying to click a single spot to focus a panorama, so any "activate panorama" events are ignored from that point as they'd be unintentional
- Formula itself is nothing special, just pythagoras as you say. Since we're only concerned with magnitude it doesn't really matter which one you subtract from the other, Maths.abs() takes care of any ambiguity there (however it appears .hypot() takes care of that anyway, now I'm running some examples. You'll see that `Math.hypot(10, 10) === Math.hypot(-10, -10)`. The 0 vertical/horizontal movement also doesn't cause issue - `Math.hypot(10, 0) === 10`

midnight-dev:
Yeah, I didn't think about it at the time, but since squares are involved, hypotenuse calculations will always produce positive results, effectively absolute.

With that out of the way, I have a couple of suggestions. 

The arbitrary threshold is good for testing, but it would be really neat if it were based on effective screen size. The higher the resolution, the more distance (in pixels) a user's cursor travels for every action compared to those with lower resolutions. So a 10 pixel movement might feel far to small of a threshold on 4K+ screens while feeling loose on 720p screens (like on smartphones). You were likely thinking about something along the same lines. It would be a nice touch, but might require some trial and error. You'll still need to find a factor/percentage to apply to the reported resolution to find a threshold that feels good, but it'll behave the same way across a wider variety of devices.

_The line numbers to follow are based on what I can see in the PR..._ As an optimization, I'd love to see the raycaster initialization come after your drag detection since your code doesn't rely on it at all. For instance, lines 221 & 222 can come just before line 235 now. You can then remove the `isDragging` bool - instead of setting it to true and later returning if it's true, just return around line 230. Then you can remove the bool check underneath that, so it's just checking for an insertion match. This should improve performance because it no longer has to position and compute rays unless the user isn't dragging 👍

PR #876, opened 2020-08-28
cleanup: remove unused buffer and dataviewer in LASDecoderWorker

zarov:
I saw some that were forgotten, this should free a little bit of memory.


m-schuetz:
Thanks!

PR #882, opened 2020-09-04
Update ReadMe with Node.js & Linux Information

midnight-dev:
Simple PR for Node version information with a note on Linux issues. This should help Ubuntu and Debian users who install node via `sudo apt-get install node` and end up with version 8 instead of a recent version.

Also, the link on the first image was broken (/wp/demo) so I changed it to the load_project example until the path no longer produces a 404 error.


midnight-dev:
By the way, don't publish PRs through VS Code using the GitHub extension because it posts the repo's PR template as the body without allowing you to modify it before publication. Thanks Microsoft 👍

PR #886, opened 2020-09-10
Remove and ignore VS Code preferences

dsroshan:


m-schuetz:
Hi, the vscode preferences are for myself, I'd like to keep them checked in.

dsroshan:
Personal IDE preferences are not a part of the project and I just thought that they shouldn't be on the repo just because they'll mess up with other contributors' preferences. I thought it'll be better if they are kept only in local working tree and untracked by git.

midnight-dev:
Aye. That's why I commented out a few in the quality of life PR #828. Always have to comment out the settings to work on anything and then uncomment before submitting a change so as to not mess with the diffs. The font is much too big for my tiny resolution. 

Stretch this across your entire screen to see a like-for-like comparison.


I don't mean to pry, but is there any reason why these settings aren't configured globally for your VS Code installation, Markus? Local project settings always override global settings, but if there are no local settings, the global ones are used on every project. Do you have to share the IDE with others, so per-project settings are necessary?

PR #909, opened 2020-10-16
Fix for Borked GPS Time / Extra Ranges

midnight-dev:
This addresses the later issue in potree/PotreeConverter#457
PDAL + Converter 2.0 created a division by zero scenario for GPS Time.
The fix is to simply check for NaN after scale & offsets are set.
We reset them to sane defaults if NaN is found.

This can be fixed in a variety of ways. I just went with something quick. Let me know if the defaults need changing. I'm leaving this open to modifications, so feel free to edit something in place if needed.

### DIFF
--- a/src/PotreeRenderer.js
+++ b/src/PotreeRenderer.js
@@ -864,6 +864,9 @@ export class Renderer {
 				let scale = initialRangeSize / globalRangeSize;
 				let offset = -(globalRange[0] - initialRange[0]) / initialRangeSize;
 
+				scale = Number.isNaN(scale) ? 1 : scale;
+				offset = Number.isNaN(offset) ? 0 : offset;
+
 				shader.setUniform1f("uGpsScale", scale);
 				shader.setUniform1f("uGpsOffset", offset);
 				//shader.setUniform2f("uFilterGPSTimeClipRange", [-Infinity, Infinity]);
@@ -988,6 +991,9 @@ export class Renderer {
 
 					let scale = initialRangeSize / globalRangeSize;
 					let offset = -(globalRange[0] - initialRange[0]) / initialRangeSize;
+
+					scale = Number.isNaN(scale) ? 1 : scale;
+					offset = Number.isNaN(offset) ? 0 : offset;
 
 					shader.setUniform1f("uExtraScale", scale);
 					shader.setUniform1f("uExtraOffset", offset);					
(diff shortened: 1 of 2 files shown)


midnight-dev:
Update: 

Set this as draft, as it doesn't work in 0, 0 cases, but it does for 1, 1. Somehow set the wrong testing values when I wrote this up.

m-schuetz:
Thanks for investigating this. I don't have a system where it doesn't work, unfortunately.

midnight-dev:
@m-schuetz Happy to help. This fixes the glitchy point cloud for me. I detailed the reasoning in the latest commit. To sum it up, a 0, 0 range for `gps-time` was making some arithmetic hit the fan due to the `initialRange` property getting set on first load in. Even though I fixed the scale & offset NaN problems, there were still problems. I think the NaN checks should remain in because it would be trivial to accidentally inject invalid parameters here.

Something to note; this is more of a hack than a proper fix. It just compares the min/max values and adds 1 to the max range if they're equal. It may not be optimal to add 1. It could be argued that a number closer to zero would be better (e.g. 0.000001) but a really low value may open us up to infamous JS floating point precision issues when subtracting & dividing.

Also, I've seen the same attribute listed as GpsTime in a different point cloud, so this hack may not work on all clouds. That said, it's simple to rename "GpsTime" to "gps-time" in the same way that "rgb" is renamed to "rgba".

I've run through some basic testing, but I couldn't think of any edge cases to try out. If you could checkout this branch to test a few point clouds and GPS time filtering, I'd appreciate it. I don't have many samples or projects that I can load.

boxel0:
Hi midnight-dev, thank you for pointing me to this page,
I have tried changing the gps-time to 1,1 but I am not getting any changes,

The point cloud is visible for some machines but not others, I have a data set which is only small that I have uploaded:
[build2.zip](https://github.com/potree/potree/files/5511384/build2.zip)

https://cloud.lasersurveying.com/potree2/examples/build2.html

Be interested to hear your thoughts

midnight-dev:
Changing it to 1, 1 would still result in division by zero, which would produce NaN, and dividing by that would give a clipping of negative infinity to positive infinity.

Instead, check out my branch & build Potree from that. That should reveal if this is really the problem or not. If you don't know how to checkout a bit branch or build the project, I can provide a built version of the JavaScript. However, it's typically much safer to build it yourself, and you'd still have to figure out how to replace the file on your system.

Alternatively, remove the `gps-time` entry from the metadata.json - this worked for an earlier project using PotreeConverter 2.0.

m-schuetz:
Thanks for the PR!

PR #921, opened 2020-11-02
Removed reference to non existing PointAttributeNames

sverbist:
Noticed that the 1.7 version removed the Potree.PointAttributeNames "enum". But at the bottom of the file there was still a reference to it. I removed this reference and I couldn't see any impact.


m-schuetz:
Thanks, seeing how hasColors is no longer referenced anywhere, I've removed it entirely.

PR #922, opened 2020-11-02
Removed duplicate static function in PointCloudMaterial.js

sverbist:
PointCloudMaterial.js contains the static function 'generateMatcapTexture' twice with identical code. I've removed the duplicate.


m-schuetz:
Thanks!

Issue #838, opened 2020-06-19
Using Measurement tools on meshes


mikepianka:
When meshes and shapefiles have been loaded into the view, mouse clicks do not register on them. Is there a way to use the measurement tools on these?

m-schuetz:
Unfortunately there isn't, the measurement tools are all only developed for point clouds.

plines:
It is possible to measure on shapefiles and other vector data.  I had to modify the code slightly to accomplish this. 
![image](https://user-images.githubusercontent.com/7026669/96746363-1945ed00-1395-11eb-99aa-76be4344ec1a.png)

smcavoy12:
@plines 

Cool! Can I ask, are you exposing the vertices of the mesh, or raycasting on to the face surfaces?

plines:
Honestly I don't recall, it's been years since I made that change.  I fixed a few things that didn't work as expected, like measure on shapefiles, fix the pointcloud tree to toggle children layers, etc.  I'd have to dig back in the code to find my solution, and this was also done on version 1.6 for reference.

smcavoy12:
Ok, thanks! I'm glad to know it's doable at least.  Our crummy solution right now is to sample point clouds from meshes using CloudCompare. If we have other threejs loaders (not the PLY or OBJ loaders Markus included) in the same scene as a pointcloud it seems to block measurement tools for everything.

We've got the children layers for GLTFs now on 1.7.2, but we're finding that it increases loading time significantly.

mikepianka:
@plines if you could find that code it would definitely be helpful. I have still been using 1.6 mainly. I would be happy to take a look into what it would take to bring that feature up to 1.7.

plines:
I think I found it now, under static getMousePointCloudIntersection, change

plines:
if(!point){
	continue;
}

let distance = camera.position.distanceTo(point.position);

plines:
to this:
if(!point){
	if (intersects.length > 0){
		point = intersects[0].point;
		distance = intersects[0].distance;
	}
	else {
		continue;
	}
	//continue;
}
else{
	distance = camera.position.distanceTo(point.position);
}

//let distance = camera.position.distanceTo(point.position);

plines:
If a pointcloud point is intersected, it takes priority.  But this logic will try to raycast to the scene's children, and use the first (closest) intersect if no pointcloud point is found.  As long as you add your vector data to the scene like this:
let shapeNode = new THREE.Object3D();
[...]

PR #932, opened 2020-11-13
Measure on vector data like meshes and shapefiles

mikepianka:
This is an example of promoting vector data like meshes / shapefiles to being first-class objects like pointclouds. Currently, vector data cannot be measured on, orbited around with Earth Controls, etc. This would be a useful feature when mixing meshes and pointclouds together in a scene, which is probably a common scenario for many users. 

PR notes:
- Tested on the develop branch using the `meshes.html` and `shapefiles.html` examples.
- Made some slight changes to `meshes.html` included in this PR to add some attributes that could be displayed in the Properties Panel when using the Point measure tool.
- If accepted, changing the static method `getMousePointCloudIntersection` to something that also mentions vectors / meshes may be more descriptive, but for the purposes of this proof of concept I decided to leave it at the moment.

Known To Do's / help needed currently:
- I am not very familiar with jsTree and could not figure out how to get the attributes in the Properties Panel to clear. For example, hovering over a point cloud with RGBA values and then onto a mesh that doesn't have them causes the RGBA values to stop updating but also doesn't clear them from the properties panel. This is probably a simple fix someone could contribute that knows jsTree. See image below.
- Most measurement tools work as expected except the Height tool, which seems like it can be a little weird when measuring a height on a mesh.
- Point shapefiles work as expected, but the line shapefiles in the example don't seem to respond to the mouse events. Maybe something to do with how three.js renders the lines? Would appreciate some input on this if someone can.

image

The related Issue is [https://github.com/potree/potree/issues/838](url). This solution builds off a helpful tip from @plines, so thanks to you!

### DIFF
--- a/src/modules/loader/2.0/DecoderWorker_brotli.js
+++ b/src/modules/loader/2.0/DecoderWorker_brotli.js
@@ -58,7 +58,7 @@ onmessage = function (event) {
 	let tStart = performance.now();
 
 	let buffer; 
-	if(numPoints == 0){
+	if(numPoints === 0){
 		buffer = {buffer: new ArrayBuffer(0)};
 	}else{
 		try{
--- a/src/modules/loader/2.0/OctreeLoader.js
+++ b/src/modules/loader/2.0/OctreeLoader.js
@@ -192,6 +192,13 @@ export class NodeLoader{
 				current.byteSize = byteSize;
 				current.numPoints = numPoints;
 			}
+
+			if(current.byteSize === 0n){
+				// workaround for issue #1125
+				// some inner nodes erroneously report >0 points even though have 0 points
+				// however, they still report a byteSize of 0, so based on that we now set node.numPoints to 0
+				current.numPoints = 0;
+			}
 			
 			current.nodeType = type;
 


midnight-dev:
I like this as a PoC. Since it can't be guaranteed that meshes would have the same attributes as point clouds, the properties panel likely needs to change. We can fake some attributes like fetching the color of the pixel underneath the cursor when meshes are clicked and position too, but what about LOD, scan angle, GPS time, point source ID, number of returns, and so on? I don't think it's possible to fake these in any useable way. It may be better to have a dummy defaults and display N/A to users. 

> - If accepted, changing the static method `getMousePointCloudIntersection` to something that also mentions vectors / meshes may be more descriptive, but for the purposes of this proof of concept I decided to leave it at the moment.

What about `getMouseGeometryIntersection`? To maintain compatibility with existing code, keep `getMousePointCloudIntersection` but have it `console.warn` that it could become deprecated and replaced. Then just forward the arguments to the geometry intersection function.

mikepianka:
Thanks! My idea for the Properties Panel was to have it dynamically show whatever relevant attributes were being hovered. So for a pointcloud, it should behave as it does. For a mesh, have it switch to displaying any attributes in the userData on the object (like name, link, snack in my screenshot above). Originally an error was thrown because the Properties Panel was always expecting integer arrays as the catch-all, so I refactored it to have it also look for userData.  [See this commit.](https://github.com/potree/potree/pull/932/commits/b3c0149a85b74cf2a14b6096da8cd777ba343dcf)

Right now, if you start a Point measure and hover over the pointcloud, then a bunny, you'll see the bunny's attributes show up in the Properties Panel but the rgba attribute gets leftover from the pointcloud (but it does stop updating since the bunny is intercepting the mouse events). Conversely, once you hover off the bunny and back onto the pointcloud, the bunny's attributes stay in the Panel but should not. I am not familiar at all with jsTree but I think some kind of "clear" method probably just needs to get called every time the Panel updates, so attributes don't get orphaned there while you are hovering the cursor around.

I like your idea for `getMouseGeometryIntersection`.

mikepianka:
Fixed the orphaned attributes bug and rolled the changes into this PR. The Properties Panel will now update correctly when hovering from a bunny to the landscape and vice versa. 

Separate PR just for the orphaned attributes bug: [https://github.com/potree/potree/pull/936](url)

m-schuetz:
Hi,

This is very cool! However, unfortunately it causes a full freeze for 11 seconds while starting the meshes.html example. 
From the screenshot below you can see that 3 bunny models are loaded and rendered, then 11 seconds freeze, and after 11 seconds, the fourth bunny model and the pointcloud appear.

image

mikepianka:
@m-schuetz hmm the meshes.html example fully loads for me in a little over 1 second with the local server. When I started this PR version 1.7.1 was the current one. Is that what you are testing it against or are you using 1.8?

image

m-schuetz:
I've tested the version in your repository, with chrome 89 on windows 10. I'll have another look.

mikepianka:
I'm using chrome 90 on windows 10. Not sure if that would make any difference.

Issue #935, opened 2020-11-16
Orphaned Attributes in the Properties Panel


mikepianka:
When hovering over pointclouds with the Point Measure tool, attributes get orphaned in the Properties Panel once the cursor moves over a pointcloud that no longer has those attributes.

Steps to reproduce the bug:
1. Start the `multiple_pointclouds.html` example
2. Start a Point Measure and hover over a lion, then the landscape
3. You will see that the normals attributes from the lion remain in the Properties Panel, but no longer update because the landscape does not have normals

![image](https://user-images.githubusercontent.com/47948499/99275333-5ddc7100-27f9-11eb-9bd6-f9d627f33259.png)

PR #936, opened 2020-11-16
Fix Orphaned Attributes in Properties Panel

mikepianka:
This PR fixes the bug noted in [Issue #935 ](https://github.com/potree/potree/issues/935). When hovering over pointclouds with the Point Measure tool, attributes would get orphaned in the Properties Panel once the cursor moved over a pointcloud that no longer had those attributes. 

Steps to reproduce the bug prior to this PR:
1. Start the `multiple_pointclouds.html` example
2. Start a Point Measure and hover over a lion, then the landscape
3. You will see that the normals attributes from the lion remain in the Properties Panel, but no longer update because the landscape does not have normals

image

The fixed behavior in this PR:
Hovering from the lion back to the landscape causes the normals attributes to disappear from the Properties Panel.

### DIFF
--- a/src/utils/Measure.js
+++ b/src/utils/Measure.js
@@ -422,6 +422,15 @@ export class Measure extends THREE.Object3D {
 					let i = this.spheres.indexOf(e.drag.object);
 					if (i !== -1) {
 						let point = this.points[i];
+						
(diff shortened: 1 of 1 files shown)


m-schuetz:
Looks good, thanks!

PR #937, opened 2020-11-18
add support for url sas tokens e.g. Azure Storage

alexz75:
I need to add SAS tokens into all requests to a private Azure Storage Blob.
Added sasToken into XHRFactory, the token added into all URLs as a query string.
To use it, should set the token before use:
`
Potree.XHRFactory.config.sasToken = "abcdef=1&qwerty=2"
Potree.loadPointCloud("https://xxx.blob.core.windows.net/potree/vol_total/cloud.js", "sigeom.sa", e => {
...`


mattmc96 (approved):
Looks good

PR #941, opened 2020-11-23
Enable query strings for EPT Loader URLs

iamscottcox:


midnight-dev:
Does this allow for an arbitrary amount & order of URL variables? Asking to see if this will effect projects that pass requests through a backend like PHP/.NET that may require tokens or hashes. This is also important if you want to pass credentials to a cloud provider like AWS.

iamscottcox:
@midnight-dev It will! We're using it to pass through `Policy` and `Key-Pair-Id` parameters to retrieve .ept files from Cloudfront.

midnight-dev:
That's great! Though could you explain the logic behind the ternary operation? It looks like a clever way to just prepend a _?_ to the start of a query string. But then it would also set a variable to itself for every URL even if it doesn't have a query string.

If that's on the mark, I can suggest a more efficient alternative, but let me know if I'm missing an obvious reason for that logic.

By the way, does this miss a query string that only has a hash mark (_#_) without a question mark? This is typically not an issue, but it could be an oversight for future client-side routers / SPA implementations that tend to use hash marks to navigate or manage state.

m-schuetz:
Hi, 

Can you give an example of how this is used? And please remove the formatting changes in LaszipLoader.js

Thanks!

PR #944, opened 2020-12-04
Update PointCloudOctree.js

bmichalowski:
Change default value and getting rid of overwriting `pickWindowSize` value for `PointCloudOctree.js`


midnight-dev:
LGTM 👍

midnight-dev (approved):
Could be constant, but so could many variables in the project. Keeping it as a variable should be safer anyway. Less room for there to be side effects.

I'm in favor of this minor cleanup, however, it's ultimately up to Markus.

m-schuetz:
Thanks for the cleanup/fix!

PR #948, opened 2020-12-13
Font size 28 is too big

sungle3737:
### **vscode setting change pull request**
- personal perfered font sizes are different, so removing that option is reasonable


potree:
please don't try to change my vscode settings.

PR #952, opened 2021-01-13
MeasuringTool, reference error

bmichalowski:
Change from `viewer` to `this.viewer` since the first is only known in the constructor.


m-schuetz:
thanks for the fix!

PR #953, opened 2021-01-14
POCLoader, replacements for GPS_TIME

bmichalowski:
Add replacements for GPS_TIME in parseAttributes func.

The lack of this entry caused an error when working with gps-time. 
I don't remember exactly where it happened but this line fixes the problem.


m-schuetz:
thanks!

PR #959, opened 2021-01-20
profile.js, remove hardcoded cords

bmichalowski:
The absence of this entry does not change the original example or the other uses of the ProfileWindow class.


marckassay:
@bmichalowski Just out of curiosity, do you know why there were hardcoded cords committed to begin with?

bmichalowski:
@marckassay In fact, without these coordinates we will receive anonymous sphere, at the default coordinates - 0, 0, 0.

Thanks for your curiosity,  in second commit, I removed this sphere completely. It probably ended up there for debugging purposes.

m-schuetz:
Yeah, most likely just debug stuff that I forgot to remove. Thanks for the cleanup!

PR #961, opened 2021-01-21
I18n zh support

yuhangch:
This PR contains two commit:

1、Correct text binding “remove_all_clipping_volumes”
2、Add i18n zh (Chinese Simple) support


m-schuetz:
Thanks for the PR!

PR #970, opened 2021-02-04
Bump version in package.json

pierotofy:
While running a build I noticed the version in package.json was out of date (latest is 1.8.0).


m-schuetz:
Thanks!

PR #989, opened 2021-03-01
Use Three.js 129 via npm packages

gchoqueux:
The PR replaces `libs/three.js` by `three npm package`.
It facilitates the management and updating of the Three library and examples.

In Potree source code, `Three` library and examples are imported, by using 
```js
import * as THREE from "three";
import * as THREE from "three/examples/jsm/lines/Line2.js";
```
In Potree examples, `Three` library is imported, by using : 
```js
import * as THREE from "../node_modules/three/build/three.module.js";
```

There may be somes reasons to use `/libs/three.JS/` ?

### DIFF
--- a/examples/360.html
+++ b/examples/360.html
@@ -38,8 +38,7 @@
 	
 	<script type="module">
 
-	import * as THREE from "../libs/three.js/build/three.module.js";
-	
+
 		window.viewer = new Potree.Viewer(document.getElementById("potree_render_area"));
 		
 		viewer.setEDLEnabled(true);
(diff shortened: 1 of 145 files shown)


m-schuetz:
Hi, thanks for the PR. It definitely makes sense but right now I'm trying to keep maintainance to the bare necessities since I'll shift focus to a Potree 2 rewrite. Also, I think there was a small manual change in three.js that was relevant for Virtual Reality functionality, so I'd have to check if that still works with the original three.js. At least one of two issues was directly addressed in three.js by now. 

So I won't merge it for now, but I'll keep it open in case I'd like to get back to it.

gchoqueux:
I update Three to version 129

PR #992, opened 2021-03-04
Fixes 991: Fix issue with latest chrome update.

### LINKED ISSUE #991: Measuring bug in Chrome 89.0.4389.72
After updating to Chome 89.0.4389.72 the measuring tools are not working.
Trying to measure distance or height will only add one marker and then stops the measuring.

timodwhit:
Fixes #991


timodwhit:
Credit goes to @Nevernat for this

m-schuetz:
Seems to work fine, at least better than now. I think there was a reason that it was explicitly set to true, but I don't know what it was and I can't figure out anything off right now.

Thanks @timodwhit and @Nevernat

Issue #1019, opened 2021-04-21
Allow loading point cloud files from private cloud servers


corylangenbach:
It would be very useful to be able to load point cloud files from private cloud storage buckets/containers, such as those provided by AWS, Azure, and GCP. A simple way to enable that would be to allow signing the URLs that are used by Potree loaders to access files.

This feature has been requested before, in a [comment on issue #756: Potree 2.0 Considerations](https://github.com/potree/potree/issues/756#issuecomment-648880845).

I will submit a PR which provides this feature.

PR #1020, opened 2021-04-21
Enabled reading point clouds from private clouds, using signed URLs.

corylangenbach:
• Add URL signing API, and use it across all cloud loaders.
    • examples/load_from_cloud.html shows how to sign URLs for Amazon Web Services, Microsoft Azure, and Google Cloud Platform, thus allowing the loading of point clouds from any of those cloud service providers.
    • examples/load_from_cloud_README.md provides notes on configuring those clouds for access.

Note that tests are not provided, since they would require credentials to access private clouds. If individuals want to test using the private clouds we used, arrangements can be made to provide read-only credentials valid for a short time.

Solution for [issue 1019](https://github.com/potree/potree/issues/1019).


midnight-dev:
This looks more robust, but I could've sworn there was a PR that allowed arbitrary url parameters for signed URLs. I use signed cookies and haven't kept an eye on that submission. Did it conflict with one of the cloud providers?

Edit: 

Never mind. The one I was thinking of only addressed EPT clouds.

corylangenbach:
We have come across a few PR’s that are somewhat similar to ours, although we believe they do not encapsulate all that ours will add to Potree. Generally, a signature is a function of the URL and the headers. The range header is needed for PotreeConverter 2.0 files (the OctreeLoader), which means that many of these approaches will not work for AWS or GCP, due to how these cloud services deal with headers. Some of the solutions seem to work for Azure, since Azure doesn’t do anything special to deal with headers.

I have listed the PR’s that we have come across below, noting the ways in which they differ from ours:

#184 :
- Should work for Azure, but cannot generalize to AWS or GCP.
- AWS and GCP signing is more complicated (i.e. needs to handle the range header), so it doesn’t follow directly from the Azure signing code.

#937 :
- Seems to work for Azure, but cannot generalize to AWS or GCP.
- AWS and GCP signing is more complicated (i.e. needs to handle the range header), so it doesn’t follow directly from the Azure signing code.

#941 :
-  Adds support for query parameters for the EPT loader, but not for any of the other loaders.

#472 :
- Addresses the new XMLHttpRequest way to get files, but the newer loaders are using fetch, instead. Our approach covers both ways of getting files — XMLHttpRequest and fetch.
- Simplistic handling that will not cover AWS or GCP signing.


Let me know if you have any more questions/concerns.

jhsct:
Hello

I implemented for S3 but had to make one change...I'm using Potree in an MVC application so instead of asking the user to enter credentials we store them off on the server per point cloud and then just provide the secured links on page load for each of the files types.  The following is the result:

`var metadatalink, octreelink, hierarchylink, folder;
async function GetCloudCredentialsSuccessFunction(result) {
        metadatalink = result.metadatalink;
        octreelink = result.octreelink;
        hierarchylink = result.hierarchylink;
        var folder = "UpstreamPotree"; // --> Later result.Folder;       
        
       var signUrl = async function(url, headers) {
           var u = "unknown";
           switch (url){
           case folder + "/metadata.json":
               return metadatalink;
               case folder + "/hierarchy.bin":
                   return hierarchylink;
               case folder + "/octree.bin":
                   return octreelink;
       }
           
       return "";
       }

Hope this helps anyone else.

I will say our performance is horrible compared to housing the data on the server and I'm not sure why.  It appears using the public buckets serves up the data fine.  I do have Transfer Acceleration Enabled on the S3 bucket so I'm not sure...

Any suggestions?

Thank you
Jeff

kevjamco:
@jhsct I just tested with S3, and am not experiencing a significant difference in performance compared to EFS. It seems completely usable. What is the region your data is stored at and where are you accessing it from?

Could you also provide some more detail on your implementation of computing the secure links server side? Thanks!

jhsct:
@kevjamco - It's been a while and I backed out the code and went back to the main released code because of the lagging.  I was in US EAST and I was in Chicago at the time so I wouldnt expect the lag.  It was probably my implementation and lack of skill and understanding on the javascript side - (most times it's me). Anyhow, on the server side, we run an MVC website and I was passing down the presigned urls because we store the Bucket Credentials in an Azure SQL instance - so easy enough to create and pass.  I could use the help if your interested...leave me a message as ecedesign,com

hamza-iqbal-ad:
> @kevjamco - It's been a while and I backed out the code and went back to the main released code because of the lagging. I was in US EAST and I was in Chicago at the time so I wouldnt expect the lag. It was probably my implementation and lack of skill and understanding on the javascript side - (most times it's me). Anyhow, on the server side, we run an MVC website and I was passing down the presigned urls because we store the Bucket Credentials in an Azure SQL instance - so easy enough to create and pass. I could use the help if your interested...leave me a message as ecedesign,com

how to load presigned url, as urls for each file is completely different so what is the way to provide these loadPointCloud function?

PR #1072, opened 2021-08-25
Fix bug when srs undefined causing point cloud not to display

OKaluza:
If info.srs is not present this is throws an error, occurring for some
point clouds in WebODM

I am not sure what exact situation leads it to have no srs, but the same check exists for the previous statement so can't really see any negatives to adding it again here. I have tested the fix and it solves the problem in the potree instance bundled with WebODM.


zoltan-mihalyi:
I have the exact same problem, and fixed the same way. @m-schuetz can you accept this PR?

chollid:
My team ran into the same issue and also implemented this exact fix.

PR #1123, opened 2021-12-16
Add catalan translation

geostarters:
just add catalan translation


geostarters:
Too many files changed

PR #1132, opened 2021-12-21
Added YZ and ZX plane area calculation for area measurement

arjan-kuiper:
By default, area measurements only work on the XY plane. This PR also calculates the area of the YZ and ZX plane and adds it to the XY calculation. This allows for plane measurements that are not necessarily on the XY plane.

### DIFF
--- a/src/utils/Measure.js
+++ b/src/utils/Measure.js
@@ -533,6 +533,8 @@ export class Measure extends THREE.Object3D {
 			let p1 = this.points[i].position;
 			let p2 = this.points[j].position;
 			area += (p2.x + p1.x) * (p1.y - p2.y);
+			area += (p2.y + p1.y) * (p1.z - p2.z);
+			area += (p2.z + p1.z) * (p1.x - p2.x);
 			j = i;
 		}
 


henriod:
Have tested your code vs one have generated from this discussion [Area of polygon in 3D plane](https://math.stackexchange.com/questions/3207981/caculate-area-of-polygon-in-3d#:~:text=Remark%20%3A%20A%20simpler,we%20will%20have) as 
```
         getArea() {
		let area = 0;
		let j = this.points.length - 1;
		for (let i = 0; i < this.points.length; i++) {
			let p0 = this.points[0].position;
			let p1 = this.points[i].position;
			let p2 = this.points[j].position;
			let a = (p2.y - p0.y) * (p1.z - p0.z) - (p2.z - p0.z) * (p1.y - p0.y);
			let b = (p2.x - p0.x) * (p1.z - p0.z) - (p2.z - p0.z) * (p1.x - p0.x);
			let c = (p2.x - p0.x) * (p1.y - p0.y) - (p2.y - p0.y) * (p1.x - p0.x);
			area += Math.sqrt(a * a + b * b + c * c);
			j = i;
		}
		return Math.abs(area / 2);
	};
```
and the is a big difference and when excel sheet is used your solution gives wrong values. I am open to any discussions
math

Issue #1167, opened 2022-03-14
Problems loading MTL


klakar:
I'm building viewers for point clouds where I experiment with including mesh models in obj-format. When including the mtl-files I get an error pointing to a "file not found"

![bild](https://user-images.githubusercontent.com/6375959/158175578-83127cf4-2fe5-4204-9345-d87d68e37472.png)

I can fix this by editing the MTLLoader.js file on row 12.

from:
_} from '../../../build/three.module.js';_

to:
_} from '../build/three.module.js';_

I don't know if this is a problem for Potree or Three.JS, but without this change I'm not able to load textures from a mtl-file for my models.

klakar:
A fix correcting the URL has been submitted as #1168

SBCV:
Thank you for pointing this out, adjusting this line allows me to run `import {MTLLoader} from "../libs/three.js/loaders/MTLLoader.js";` without errors as well.

Are you able to share a simple example, how to use the `MTLLoader` with potree? I tried to incorporate [an example of the three.js documentation](https://github.com/mrdoob/three.js/blob/master/examples/webgl_loader_obj_mtl.html) without success :o

SBCV:
Nevermind, was a stupid mistake at my end ...

Extending the `meshes.html` example with the following produced the desired results
```
...
import {MTLLoader} from "../libs/three.js/loaders/MTLLoader.js";
...
new MTLLoader()
.setPath(`${Potree.resourcePath}/models/my_model/`)    // do not forget the last slash
.load( 'textured.mtl', function ( materials ) {
	materials.preload();
	new OBJLoader()
		.setMaterials( materials )
		.setPath(`${Potree.resourcePath}/models/my_model/`)  // do not forget the last slash
		.load( 'textured.obj', function ( object ) {
			// Do not remove these settings, since they substantially improve the navigation. 
			object.position.set(589871.587, 231528.213, 725.634);
			object.scale.multiplyScalar(500);
			object.rotation.set(Math.PI / 2, Math.PI, 0)
			viewer.scene.scene.add( object );

		}, onProgress, onError );

} );
```

PR #1168, opened 2022-03-14
Update MTLLoader.js

klakar:
Fixing File Not Found error when trying to load MTL files (#1167)

(compare with OBJLoader.js file)


SBCV:
Would be nice if someone could merge this fix - without it it is not possible to display MTL files

PR #1206, opened 2022-07-01
Fix error loading EPT data

sketchpunk:
I used entwine to create a few point clouds in the EPT format. When using potree to view the pointcloud there was an error that prevented anything from rendering. It was a very simple fix by including a undefined check to an if statement. This might of been missed since the if statement before the one I changed also had a undefined check for info.srs.


m-schuetz:
Thanks for the PR!

PR #1234, opened 2022-09-14
Added back the rendering of NavigationCube and fixed its raycasting

Neralem:
The navigation cube button in the potree toolbar did not bring up the navigation cube because it was commented out in the renderer. I've added it back in and it showed up but behaved strange. I've found out that the hit testing when clicking on it with the raycaster did not take into account that the render area may be not in the most topleft corner of the browser viewport like in my case.

So I've asked the canvas which fired the onMouseDown event to give me it's x and y offset relative to the browsers viewport topleft corner:
`var viewportOffset = event.toElement.getBoundingClientRect();`

and added it to the calculation:
```
mouse.x = (event.clientX - viewportOffset.x) - (event.toElement.clientWidth - this.width);
mouse.y = (event.clientY - viewportOffset.y) - this.yOffset;
```

Notice that `this.yOffset` is also added by me and allows to shove the NavigationCube further down. It is 0 by default so no impact if you don't change that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,4 @@ libs/CesiumUnminified
 /pointclouds/morro_bay_converted
 /pointclouds/helimap_epalinges
 pointclouds/F
+/.vs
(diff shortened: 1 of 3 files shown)


swell-d:
thanks!

PR #1328, opened 2023-05-29
Allow using arbitrary skyboxes

gzotti:
- use URL parameter 'skybox' naming customized skybox
- avoid caching textures to allow refreshing skybox
- use URL parameter 'skyboxNorth' to allow corrective azimuth rotation

Rationale: Combine a pointcloud with a sky background that has actually important scene content. 


mgalvi:
I will give it a try! Thank you for sharing

PR #1367, opened 2023-10-06
Camera animation improvements.

MyosQ:
Replacing Duration with Speed (m/s).

I found it tricky to use the camera animation, since you have to specify the duration of the animation.  The animation will then move faster or slower depending on the distance it has to move. Instead, I think it is more intuitive to specify the speed directly.

In this update, you specify speed instead of duration. In addition, I added
* repeat - if the animation should loop or not.
* stop - a method for stopping the animation.
* acceleration - specified in m/s2. This make the animation smoother. Instead of going from 0 to the given speed, it accelerates up to that speed.

I updated the sidebar panel correspondingly.

image

### DIFF
--- a/src/modules/CameraAnimation/CameraAnimation.js
+++ b/src/modules/CameraAnimation/CameraAnimation.js
@@ -41,11 +41,15 @@ export class CameraAnimation extends EventDispatcher{
 		this.node.add(this.frustum);
 
 		this.name = "Camera Animation";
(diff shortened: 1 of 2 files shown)


m-schuetz:
Thanks for the PR. That change makes perfect sense, only issue I'm concerned about is backwards compatibility. Users who already made animations would have to migrate somehow.

Issue #1065, opened 2021-08-12
Add COPC Support


timodwhit:
A new format is coming out (Cloud Optimized Point Cloud (COPC)) https://github.com/copcio/copcio.github.io/ 

It is just a LAZ file but it is LAZ 1.4 and contains a hierarchy in the eVLR that matches the EPT hierarchy format (roughly). 

The spec is being finalized and it would be nice to have LAZ1.4/COPC support for potree visualization.

timodwhit:
@connormanning with your work here: https://github.com/connormanning/copc.js are you planning to tackle this? Or would you like some help? It seems fairly straightforward but what are you thinking for integration?

connormanning:
@timodwhit I have no imminent plans to do this - I am currently adding COPC support to EPT Tools, which is going to help me fill out my copc.js API footprint with the rest of the things you'd need to do the Potree task.  For example voxel key traversal and bounds bisection.  I already have these things implemented elsewhere, but I need to integrate them into my copc.js API.  I would expect to be done with that shortly - this week for sure.  After I've done this, I think you are correct that this task would be fairly straightforward.

timodwhit:
Good Stuff. Thanks @connormanning 

@m-schuetz You requested that we try to segment the COPC work into it's own piece as much as possible, and Connor's work does a lot of that. Do you have any thoughts on how you would like this integration done to ease the migration for 2.0? Should we use the EPT loader as an example or do you have another example we should use as a basis? The major concerns I have right now are library management and the switch to say "Load COPC under these circumstances."  

As the COPC spec does not require a certain name of the file to be a COPC file (understandably), we could likely have the requirement for potree or we could check against a header if laz file.

timodwhit:
@connormanning just wanted to check and see how the ept-tools work was going. I got derailed on other work but should be tackling this next week or so. 

Thanks!

nejcd:
Any new development on this topic? 

I am interested to work on this feature or support the work.

HjalmarCarlson:
Also interested to know where this is and if support is needed to move this forward.

timodwhit:
The work here was closed out https://github.com/connormanning/copc.js/pull/1, once there is a web build of the copc js llib that could be used in potree, then the copc loader would need to be copied (essentially) from the ept loader using those methods.

connormanning:
[...]

Issue #756, opened 2020-01-15
Thinks to be aware of


m-schuetz:
Some things to be aware of for future work on Potree.

- Handling large number of individual point clouds
- Better control of near/far plane. Perhaps allow custom values as well. 

## Attributes
- Handle arbitrary number of attributes, and composites of arbitrary combinations of attributes
- Handle different classes of attributes: scalar (intensity), vector of scalars (normals), RGB values, enumerations (classification, source ID, ...), double values (GPS Time)

## File Format
- Single file for octree data, multiple files for metadata? e.g. cloud.json, hierarchy.data, octree.data, ...

m-schuetz:
Display current projection. E.g. in Measurement Panel, show projection along with the coordinates.
, as suggested by @mriedo
![image](https://user-images.githubusercontent.com/6705073/72608435-f2005200-3922-11ea-9e5f-3cc4673e39db.png)

silqwer:
Please support the latest three.js version.

jhlwscom:
Incredibly keen to see the new file format changes go in.

Our current project involves hosting 15-30 billion point clouds in azure blob storage, and the amount of files generated makes querying the contents of the container quite problematic. Also worth noting that a nice method to add a Signed Access Signature / token to a URL would be very helpful. 

I.e latest upload was 2095768 individual files.

Also, support for multiple point clouds/layers will be incredibly useful in cases where users need to compare two clouds (i.e same  are different time)

mriedo:
Multiple pointcloud/layers is possible. You can see an example here : https://sitn.ne.ch/lidar/
with three surveys 2001, 2010 & 2016 ...
You can choose different colors for the layers and see the differences

jhlwscom:
Ah, I wasn't very clear. There's different annotations and things per layer, associated with each pointcloud. I.e cloud from 2016 has different annotations to 2017. 
Currently we use subgroups in annotations, which works well enough.

daviddc86:
Hello
I need some help.
I am trying to add the pointlcoud 2.0's to a project with AutoDesk Forge Viewer that uses THREE 71.

I read the files well but the material is different for the THREE version and I get a figure with triangles and many blanks.

For the previous pointclouds @petrbroz made a demo, but now I can't find any information.
![image](https://user-images.githubusercontent.com/4498677/121161339-2ee92380-c84d-11eb-9afd-87ee261a9cb3.png)

timodwhit:
[...]

PR #1381, opened 2023-11-29
Add initial COPC and LAS 1.4 support

connormanning:
This PR adds support for [COPC](https://copc.io/) and LAS (and LAZ) 1.4 data, including EPT data which uses LAS 1.4 as its data storage.  Some sample COPC data can be found in the `Data` tool [here](https://viewer.copc.io/).  Some key things in the code are as follows.

This work was accomplished by [Hobu Inc](https://hobu.co/) during the [OSGeo Community Code Sprint 2023](https://wiki.osgeo.org/wiki/OSGeo_Community_Sprint_2023).

## Examples
After `npm start`, some very small embedded test datasets are here:
- COPC: http://localhost:1234/examples/copc.html
- EPT with LAZ 1.4 encoding: http://localhost:1234/examples/ept_laz_14.html

And you can link out to more substantial data:
- [SoFI Stadium COPC](http://localhost:1234/examples/copc.html?r=https://s3.amazonaws.com/hobu-lidar/sofi.copc.laz&c=intensity%20gradient) (360M point COPC, ~2GB)
- [Hobbs NM EPT](http://localhost:1234/examples/ept.html?r=https://na-c.entwine.io/coh/ept.json&c=rgba) (24B point EPT)

## copc.js
Lots of functionality has been offloaded to the [copc.js](https://github.com/connormanning/copc.js) library which has utilities for reading LAS/LAZ/COPC/EPT data and is fairly well unit-tested (PRs graciously accepted).  Parsing LAS headers, loading hierarchies, decompressing point data, and extracting the attributes via views are all accomplished via copc.js.  Since copc.js is distributed as an ES module, I used [copc-umd](https://github.com/connormanning/copc-umd) to generate a UMD bundle for use by this project.

### Geometry
The COPC/EPT geometries are quite similar, and the identical functionality has been extracted into a [BaseGeometry](https://github.com/connormanning/potree/blob/f7ac011bedd4910ee2f70c1f240671161962a413/src/PointCloudEptGeometry.js#L50).  The [COPC/EPT](https://github.com/connormanning/potree/blob/f7ac011bedd4910ee2f70c1f240671161962a413/src/PointCloudEptGeometry.js#L90-L161) geometries parse metadata into a common format and create their point data loaders, and specify the mechanism for loading a hierarchy page: these are just slightly different methods from copc.js.

The [GeometryNode](https://github.com/connormanning/potree/blob/f7ac011bedd4910ee2f70c1f240671161962a413/src/PointCloudEptGeometry.js#L163) for both formats is the same since the differing functionality is encapsulated in the above.  Perhaps it should have a better name since it applies to both formats.

### Loader
The [EptLaszipLoader](https://github.com/connormanning/potree/blob/f7ac011bedd4910ee2f70c1f240671161962a413/src/loader/ept/LaszipLoader.js#L3) has been updated to use the most recent laz-perf/WASM (via copc.js).  The decompression of the actual laszip data has also been moved out of the loader ([previously it was here](https://github.com/potree/potree/blob/develop/src/loader/ept/LaszipLoader.js#L58-L78)) and into the web worker.  The [CopcLaszipLoader](https://github.com/connormanning/potree/blob/f7ac011bedd4910ee2f70c1f240671161962a413/src/loader/ept/LaszipLoader.js#L44) is quite similar except that it passes a compressed range of data from a COPC file rather than a full laszip file.

### Worker
The web worker where laszip data is [decompressed](https://github.com/connormanning/potree/blob/f7ac011bedd4910ee2f70c1f240671161962a413/src/workers/EptLaszipDecoderWorker.js#L19-L24) and [extracted](https://github.com/connormanning/potree/blob/f7ac011bedd4910ee2f70c1f240671161962a413/src/workers/EptLaszipDecoderWorker.js#L95) into buffers looks very different.  However this wasn't purely for stylistic reasons: the attributes are pulled out more generically to allow seamless support for extra-byte attributes (compare with [previous code](https://github.com/potree/potree/blob/develop/src/workers/EptLaszipDecoderWorker.js#L34-L52) which has a strict set of attributes).  However this code is not currently adding extra-byte data to the output at the moment, though they could trivially be added to the map and extracted in the loop the same way as the existing attributes.

### Bug?
I may have introduced a bug where nodes are destroyed with `dispose()`, but the renderer still tries to render them.  This would cause a crash because the node is missing its `geometry` after disposal.  I made some attempts to fix it and now I'm having trouble reproducing it, but I'm not fully convinced it's fixed.  I've added this [console.log](https://github.com/connormanning/potree/blob/f7ac011bedd4910ee2f70c1f240671161962a413/src/PotreeRenderer.js#L860), which is triggered immediately before crashing.  Usually this would occur when zooming out, and in my experience it has always been on a level 1 node like `1-1-0-1` or similar.  So I'm not positive that this PR is production ready, but should be 99% of the way there.

## See also
- Completes #1065
- Relevant to #756


jo-chemla:
Incredible to see these two pieces come together! The Potree viewer being one of the most used implementation for tiled pointclouds, and COPC being the open-standard made to store and stream massive tiled pointclouds. 
Note @m-schuetz this might be a great direction when you mentioned supporting extra per-point attributes for Potree-Next. 

Also note: PotreeConverter v2 is also the fastest pointcloud tiler I've come across - would be interesting to redo the benchmark. Would also be great if pdal copc conversion could use the same kind of optimization - or if potreeconverter could have an option to export to COPC.

m-schuetz:
Awesome, thanks for this work! It will take me a couple of days to dig through it. Too bad I wasn't able to be at the code sprint, in my home-town of all places.

m-kuhn:
Great work, thank you so much for this @connormanning, I didn't actually realize you were working on this when we talked back at the codesprint.

I just tested it on a 27GB COPC and it worked perfectly fine.

m-kuhn:
Note: seems it works if I load it with `npm start` but not if served as static files from a server, in this case it will just try to download the complete file apparently. Note: It may also be me doing something wrong..

Rdataflow:
@m-kuhn just guessing, maybe those commands didn't run successfully on your box?
```
npm run build
npm run postinstall
```
nb. on my box static files seems to work

m-schuetz:
Nice work, haven't encountered any issues so far so I'd be ready to merge. Just wondering about one thing:

The COPC example you've linked loads with 4 concurrent http requests, while the EPT example only loads with 1 http request at a time. Is this a difference in some settings in the code? Or could it be due to server configs? I think I've had cases where loading data from a server from a different domain limited the amount of concurrent requests to 1. 

[SoFI Stadium COPC](http://localhost:1234/examples/copc.html?r=https://s3.amazonaws.com/hobu-lidar/sofi.copc.laz&c=intensity%20gradient) (360M point COPC, ~2GB)
image

[Hobbs NM EPT](http://localhost:1234/examples/ept.html?r=https://na-c.entwine.io/coh/ept.json&c=rgba) (24B point EPT)
image

Rdataflow:
> The COPC example you've linked loads with 4 concurrent http requests, while the EPT example only loads with 1 http request at a time. Is this a difference in some settings in the code? Or could it be due to server configs? I think I've had cases where loading data from a server from a different domain limited the amount of concurrent requests to 1.

just guessing: could be related to server providing http/2 connection (multiplexing)

m-kuhn:
Indeed I had not run the `npm run postinstall`, thank you.

It loads and shows fine, but I still experience an "infinite load" setting, where data continues to be downloaded (visible in the developer tab).

Rdataflow:
> It loads and shows fine, but I still experience an "infinite load" setting, where data continues to be downloaded (visible in the developer tab).

odd to observe in your implementation the request has `Range: bytes=0-` 👀  ... thus consequently it will load the entire file

Rdataflow:
> > It loads and shows fine, but I still experience an "infinite load" setting, where data continues to be downloaded (visible in the developer tab).
> 
> odd to observe in your implementation the request has `Range: bytes=0-` 👀 ... thus consequently it will load the entire file

true, I can reproduce it (most of the times). @connormanning this looks to me like a bug

connormanning:
Thanks for the feedback, all.  I will take a look at:
- single-threaded EPT
- full file download request when run from the compiled version, rather than the dev version

Neither of these are the expected behavior so they both sound like bugs to me.

Rdataflow:
> The COPC example you've linked loads with 4 concurrent http requests, while the EPT example only loads with 1 http request at a time. Is this a difference in some settings in the code? Or could it be due to server configs? I think I've had cases where loading data from a server from a different domain limited the amount of concurrent requests to 1.

Situation
- https://s3.amazonaws.com/hobu-lidar/sofi.copc.laz is served via `HTTP/1.1` (handles 1 request per connection)
- https://na-c.entwine.io/coh/ept.json is served via `HTTP/2` (allows multiplexing multiple requests through 1 single connection)

Investigation
- Adding a `HTTP/2` capable reverse proxy in front of the former file shows a single connection pattern too (as expected for HTTP/2 multiplexing)
image

Steps to reproduce
run HTTP/2 capable nginx server with this snippet
```
    location /hobu-lidar/sofi.copc.laz {
        proxy_pass https://s3.amazonaws.com/hobu-lidar/sofi.copc.laz;
    }
```

Conclusion
- single or multi connection pattern is only related to HTTP/1.1 or HTTP/2 protocol support on serverside

cc @connormanning

Rdataflow:
> It loads and shows fine, but I still experience an "infinite load" setting, where data continues to be downloaded (visible in the developer tab).

This reproduces in both: dev mode (npm start) as well as in static served versions.
The infinite loading is observed in chromium based browsers (chrome, edge) with `range: bytes=0-` but is being blocked in firefox trying `range: bytes=0--1`
It seems to occur only with a single COPC prepared by @m-kuhn - maybe an unexpected file structure not being caught?

cc @connormanning

m-schuetz:
> single or multi connection pattern is only related to HTTP/1.1 or HTTP/2 protocol support on serverside

I've once had the same data set on the same server serve with 1 or 6 parallel requests. 

* 6 parallel requests if the potree page was on the same domain as the data
* 1 parallel request if the potree page was on my local PC and loading from that server. 
* 6 parallel requests if the potree page was on my local PC and loading from a local webserver. 

So I was wondering if your results may be due to the configuration of the proxy limiting CORS requests, rather than the difference between http 1 or 2?

connormanning:
Yes, this must be something to do with the CloudFront proxy.  If you use the direct S3 link to the same data [here](http://localhost:1234/examples/ept.html?r=https://s3.amazonaws.com/na.entwine.io/coh/ept.json&c=rgba), the pattern looks as expected.

One note: even with the proxied version, I think the data *is* being fetched concurrently.  It shows up as a single line in the overview, but when I enabled network throttling so the requests were very slow, I could see that requests were actively downloading data concurrently.  Both while the requests were occurring, and also afterward I could see that the timestamps for the download portion of various requests were overlapping.

So I don't think there's an issue here.  I'll look at the "whole file download" issue next.

connormanning:
Fixed [here](https://github.com/potree/potree/pull/1381/commits/81bd6ff534e9dd28e94233e703ef226101b80214).  COPC allows point data nodes to be empty, but the code was not handling this.  When this happens, the node info looks like:
```
pointCount: 0
pointDataOffset: 0
pointDataLength: 0
```

Of course, in this case we do not want to try to fetch/decompress the point data which does not exist, but the previous code was computing a range request from `0` to `-1` in this case since range requests are inclusive so we must subtract 1 from the end of the range.

So both issues mentioned here by users should be resolved - feel free to post if anything else comes up.

Rdataflow:
I can confirm the whole file download is gone with your fix 👍

klakar:
Got this to work on a local web server. "Old" ept.json and .copc.laz files work the same without major issues.
Sometimes the GUI loads without point cloud, but a page refresh usually solves that.

On mobile (Android) however, I can't get any point clouds to show... (not with COPC or Ept)
I've tried on phone and tablet, but on the same web server, so it might be the server that has some issues?
"Stock" Potree with ept works fine on mobile...

(On iOS it shows, but no colors. Chrome on Android will show point cloud in VR but not in normal view...)

m-schuetz:
Gave it a try on my Android Phone (Galaxy S20FE; Opera Browser) and it works, so I'll merge the PR!

Thanks @connormanning and @hobu !

PR #1383, opened 2023-11-30
fix: normal attribute buffer size for decode octree.bin

feelgom:
# Motivation for the Task:
The default Potree converter 2.0 supports the LAS format as input by default. However, I needed to convert point clouds with normal vectors. I decided to use the Draco format for conversion, and thus, I added the Draco Loader to the Potree Converter. Unfortunately, in the Potree Viewer, the Potree converted from Draco is unable to load normal data correctly.

# Details of the Task:
In src/modules/loader/2.0/OctreeLoader.js: line 86, geometry.attribute.normal's itemSize is set as 3. However, when decoding the octree.bin file in src/modules/loader/2.0/DecodeWorker.js and src/modules/loader/2.0/DecodeWorker_brotli.js, the normal attribute buffer size is set as pointsNumber*(typeSize). I believe it should be set as pointsNumber*(typeSize)*(itemSize), so I added an else if clause to set the correct normal buffer size.

# Future Work:
However, I'm not sure if the Brotli loader code is written to function properly. 
I request a review.
Thanks.


feelgom:
Hi @m-schuetz 

I hope this message finds you well. I wanted to kindly follow up on this PR which has been open for over a year now. 

This PR addresses an important issue with normal vector loading in Potree when converting from Draco format. Specifically, it fixes the buffer size calculation for normal attributes in the DecodeWorker, which currently causes incorrect normal data loading. The fix involves properly accounting for the itemSize (3) when calculating the normal attribute buffer size.

Given that this affects the core functionality of normal vector visualization for Draco-converted point clouds, I believe this would be a valuable improvement for users working with normal data in their point clouds.

Would it be possible to get a review when you have a chance? I'm particularly interested in feedback on the Brotli loader implementation to ensure it's handling the changes correctly.

Thank you for your time and consideration!

PR #1392, opened 2023-12-20
tidy up vertices for computeTransformedBoundingBox

michael-erskine:
Removed duplicate vertex and re-ordered remaining vertices to something meaningful.


m-schuetz:
Thanks for the PR! Looks good.

PR #1427, opened 2024-04-10
annotation tool: Fix `viewer` not being in scope.

nh2:
Like the other functions do.

It only works on the examples because they put `viewer` into the `window` scope.


m-schuetz:
Thanks for the fix!

DroneSurveyCR:
Discount applied

PR #1448, opened 2024-07-03
Add Virtual Point Cloud loader

vanlankveldthijs:
Add a loader for VPC (virtual point cloud) files.

* Datasets referenced in the VPC files that are COPC files are added as individual scenes.
* VPC files should also support LAS/LAZ files, but it seems like the las/laz loader supplied with potree does not work.
* This loader (temporarily) disables support for Veesus 4D Arena (vpc) files. It is unclear how to differentiate these from Lutra/North Road/Hobu VPC files.
* Example included for loading 'small' VPC file.


Rdataflow:
@m-schuetz IIUC this will help to view arrays of COPC.LAZ in potree - which would be an awesome step to take potree to the next level and enable it for use cases with region/country wide patches of COPC.LAZ, not?

PR #1461, opened 2024-08-28
include Z values for shapefiles

jrjdavidson:
I generalised the shapfile loader so that it includes z-values if they are present on the file- they were silently in the current version. Also added multi-point and multi-line while at it.


mlinds:
This would be a very useful addition - I hope this PR is reviewed and merged to the main branch.

trns1997:
@mlinds @jrjdavidson any idea to when this will be merged?

jrjdavidson:
Sorry I can't help you here. My personal opinion is that it might not be merged as it might be overwritten by a future update to shapeloader.js as pointed out in #535.
If you did want to use this implementation , all you would need is to copy the two files changed into your project!

